PR #7, opened 2020-02-11
Fix lowercase def:  __64bit__ to __64BIT__

kgerheiser:
The defs in CMake somehow got turned lowercase to \_\_64bit\_\_ when it should be \_\_64BIT\_\_


climbfuji:
@kgerheiser thanks, will merge, but did this actually cause any problems downstream? Or just warnings?

kgerheiser:
I didn't see any problems, I just noticed that it didn't match the actual ifdefs

climbfuji:
Thanks, then I don't need to update the tag.

PR #11, opened 2020-03-01
clean up

arunchawla-NOAA:


aerorahul:
`src/grib2.doc` contains the documentation of this library.  Part of it should be moved to the `README.md` and other part should be turned into a proper documentation inline with the code so that one can build it with `doxygen` or `sphinx`. The later can be done after the public release, but it would be nice to have that information handy.

PR #14, opened 2020-06-17
Remove dependency on bacio

kgerheiser:


BoiVuong-NOAA (approved):
Looks good

edwardhartnett:
Well it would be good to know why this was done, because there is still code from bacio in g2. ;-)

BoiVuong-NOAA:
I think because user has never known this routine existing in G2 library.

edwardhartnett:
Yet the routine was not removed, just the linking to bacio.

How do we know the function is not used?

BoiVuong-NOAA:
I do not know the G2 has been introduced to public since 2008.

edwardhartnett:
I don't really mind about the "public" using it. I care about any NOAA users using it...

BoiVuong-NOAA:
I do not know.

PR #50, opened 2021-05-17
Remove getidxes.f

### LINKED ISSUE #49: Repeat declarations of getidx causes bug
There are two files: `getidx.f` and `getidxes.f`

And they both declare the same subroutine `getidx`, which are nearly exactly the same.

However, this line is different between them:

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/25457ac15b6ee80adc2308d598ac9c40028c6001/src/getidxes.f#L95

and

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/25457ac15b6ee80adc2308d598ac9c40028c6001/src/getidx.f#L98

One checks if `lugb` is greater than `100` and the other checks if it's greater than `9999`.

Depending on the order of compilation (or maybe order of the symbols in the library?) either could be picked up. 

This causes a bug as reported here (https://github.com/NOAA-EMC/hpc-stack/issues/212) when `lugb` equals 201 the one version returns an error, but the other is successful.

@BoiVuong-NOAA do you know why both of these files declare `getidx`, and can you confirm that `getidxes.f` is the correct one?

kgerheiser:
This file contains a duplicate declaration of the subroutine getidx.

Fix #49


BoiVuong-NOAA (approved):

BoiVuong-NOAA:
Kyle, 
You are right. When I compiled g2 v3.1.0 library, I did not compile the routine getidex.f as your Cmake.  It included in your list_of_files.cmake.

PR #52, opened 2021-06-07
Remove 'save' attribute on idxlist and deallocate entry if it already exists

kgerheiser:
This fixes a bug reported by Biju where passing the same file unit (LUGB) results in an error.

I ran a diff between what was in `getidxes.f` and what is currently in `getidx.f` and changed `getidx.f` to match.

Looks like it checks if an entry with the file unit (`LUGB`) already exists and deallocates it before proceeding, but in `getidx.f` this check wasn't there.

I also removed the `save` attribute from `idxlist`.

@BoiVuong-NOAA what do you think of these changes? I'm not really sure why the one version had this check, or why `idxlist` was marked as `save`.

### DIFF
--- a/src/getidx.f
+++ b/src/getidx.f
@@ -72,7 +72,7 @@ C$$$
          character(len=1),pointer,dimension(:) :: cbuf
       END TYPE GINDEX
      
-      TYPE(GINDEX),SAVE :: IDXLIST(10000)
+      TYPE(GINDEX) :: IDXLIST(10000)
 
       DATA LUX/0/
 C - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
@@ -111,6 +111,11 @@ C  DETERMINE WHETHER INDEX BUFFER NEEDS TO BE INITIALIZED
          IDXLIST(LUGB)%NNUM=0
          LUX=0
       ENDIF
+
+      IF (ASSOCIATED(IDXLIST(LUGB)%CBUF)) then
+         DEALLOCATE(IDXLIST(LUGB)%CBUF)
+      end if
+
       IF (LUGI.LT.0) THEN      ! Force re-read of index from indexfile
                                ! associated with unit abs(lugi)
          IF ( ASSOCIATED( IDXLIST(LUGB)%CBUF ) ) 


BoiVuong-NOAA:
If you remove save from idxlist, then, the index will be re-generated each time you open an grib file.  It will work for same file unit.

BoiVuong-NOAA (approved):

kgerheiser:
What are the implications of regenerating the index? Should I re-add the `save`?

BoiVuong-NOAA:
No, You do not need it.   In the past, we have tried to save some 
resource by saving the index file. Now, We have faster machine. 
Therefore, we do not need save the index file.   The index file will be 
generated each time we read the grib file.

We need to test with some programs in operation to make sure they work.

kgerheiser:
Well, the code in this subroutine was never called because of the conflicting declarations of `getidx`. Now it matches the (working) version that took precedence in `getidxes.f` before it was removed.

BijuThomas-NOAA:
Let me know when the g2 is built with these fixes.

BijuThomas-NOAA:
Has g2 built with last fixes(c51c8f0)? Also wondering when  g2/3.4.2 available on WCOSS-Cray.

kgerheiser:
I will do 3.4.3 release and then we will install it and @Hang-Lei-NOAA can install it on WCOSS-Cray

Hang-Lei-NOAA:
Will install it on hera and wcoss

kgerheiser:
3.4.3 has been tagged

BijuThomas-NOAA:
> Will install it on hera and wcoss
Has g2/3.4.3 installed on WCOSS-CRAY?
I did not see it here: /usrx/local/nceplibs/NCEPLIBS/cmake/install/NCEPLIBS-v1.3.0/modules

kgerheiser:
@Hang-Lei-NOAA was it installed on Wcoss-Cray?  It has been installed on other systems.

Hang-Lei-NOAA:
Just add it on Luna.
Same one was on surge.

edwardhartnett:
Is there an issue for this? If not, can someone add one?

kgerheiser:
https://github.com/NOAA-EMC/NCEPLIBS-g2/issues/63

BoiVuong-NOAA:
Kyle, 
We need to put back the SAVE and check for condition of grib index file.  We will not allow user to open the same unit file (LUGB).
If we do not have SAVE, the routine getgb2 will continue to generate the index file for each time its call to search for fields in grib2 file.

PR #65, opened 2021-07-12
Feature/docus add doxygen for code addgrid.f

### LINKED ISSUE #59: doxyfile needs to be updated and some settings turned on
Same as https://github.com/NOAA-EMC/NCEPLIBS-w3emc/issues/81.

There are only about 65 files, so if we can do 5 a day, we can have everything done in 3 weeks. I did one today and it took about 10 minutes.

At the suggestion of @arunchawla-NOAA this should be done before the documentation for w3emc.

Hang-Lei-NOAA:
Add doxygen one by one.

### DIFF
--- a/src/addgrid.f
+++ b/src/addgrid.f
@@ -2,11 +2,11 @@
 !>    @brief This subroutine packs up a Grid Definition Section
 !>    (Section 3) and adds it to a GRIB2 message.
 !>    @author Gilbert ORG: W/NP11 @date 2000-05-01
-!>
+!>     
 !>    This routine is used with routines gribcreate(), addlocal(), addfield(),
 !>    and gribend() to create a complete GRIB2 message. Subroutine
 !>    gribcreate must be called first to initialize a new GRIB2 message.
-!>
+!>     
 !>    @param[inout] cgrib Character array to contain the GRIB2 message
 !>    @param[in] lcgrib Maximum length (bytes) of array cgrib.
 !>    @param[in] igds Contains information needed for GRIB Grid Definition Section 3.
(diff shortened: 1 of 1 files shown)


src/addgrid.f
```diff
@@ -1,67 +1,47 @@
+!>    @file
+!>    @brief This subroutine packs up a Grid Definition Section
+!>    (Section 3) and adds it to a GRIB2 message.
+!>    @author Gilbert ORG: W/NP11 @date 2000-05-01
+!>
```
edwardhartnett: Need to doocumentation blocks.

Basically, end the first one, then a blank line, then start a new one.

src/addgrid.f
```diff
@@ -1,67 +1,47 @@
...
+!>    for which optional grid points are defined.
+!>    @param[out] ierr Error return code.
+!>    - 0 = no error
+!>    - 1 = GRIB message was not initialized.  Need to call routine gribcreate first.
+!>    - 2 = GRIB message already complete.  Cannot add new section.
+!>    - 3 = Sum of Section byte counts doesn't add to total byte count.
+!>    - 4 = Previous Section was not 1, 2 or 7.
+!>    - 5 = Could not find requested Grid Definition Template.
+!>
+!>    REMARKS: Note that the Local Use Section ( Section 2 ) can only follow
```
edwardhartnett: Change REMARKS to use @note tag instead:

`@note The Local Use Section ( Section 2 ) can only follow...`

src/addgrid.f
```diff
@@ -1,67 +1,47 @@
...
+!>          Section 1 or Section 7 in a GRIB2 message.
+!>
```
edwardhartnett, line 41: Subroutine @author and @date tag needed here. Just a repeat of the one used in the file doxygen block.

edwardhartnett (approved):
Much closer!

edwardhartnett:
Also please change the PR title to include the name of the file you are fixing.

Just using the branch name as a PR title is not very helpful - we need more detail, so we can easily tell what each PR is doing.

src/addgrid.f
Hang-Lei-NOAA: addressed, Thanks!

Hang-Lei-NOAA: addressed, Thanks!

src/addgrid.f
Hang-Lei-NOAA: addressed, Thanks!

src/addgrid.f
Hang-Lei-NOAA: add it, Thanks!

Issue #55, opened 2021-07-05
doxygen cleanup needed


edwardhartnett:
Lot's of doxygen work is needed.

edwardhartnett:
This issue should not be closed until all files have been properly doxygenated.

edwardhartnett:
This should not be closed until all doxygen ghanges are done.

Hang-Lei-NOAA:
Yes, you are right, I will reopen it after the pull request next time. It
was automatically closed.

edwardhartnett:
When I say "Part of #N" then the issue is not closed. When I say "Fixes #N" the issue is closed...

Hang-Lei-NOAA:
The system says that you may only linked with maximum 10 pull request.

Hang-Lei-NOAA:
All source code in g2 library has been added doxygen docs.
Tested on without warnings.
Shown on
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/files.html
This issue is finished, closed and continue on other issues.

edwardhartnett:
Nice work @Hang-Lei-NOAA !

PR #71, opened 2021-07-14
Remove ORG and add full author name, Part of issue #55

Hang-Lei-NOAA:
This is part of issue #55 
According to Edward's decision, the ORG info is removed.
The Full name of author is added for all doxygen added code files.

### DIFF
--- a/src/enc_jpeg2000.c
+++ b/src/enc_jpeg2000.c
@@ -56,9 +56,9 @@
  * 1 = try increasing number of guard bits otherwise, no additional options
  * @param[in] jpclen  - Number of bytes allocated for new JPEG2000 code stream in outjpc.
  * @param[in] outjpc - Output encoded JPEG2000 code stream
- * @return 0 = Length in bytes of encoded JPEG2000 code stream
- *         -  -3 = Error decode jpeg2000 code stream.
- *         -  -5 = decoded image had multiple color components.
+ * @return 0 successfully decode JPEG2000 code stream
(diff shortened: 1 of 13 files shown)


src/enc_jpeg2000.c
```diff
@@ -56,14 +56,14 @@
  * 1 = try increasing number of guard bits otherwise, no additional options
  * @param[in] jpclen  - Number of bytes allocated for new JPEG2000 code stream in outjpc.
  * @param[in] outjpc - Output encoded JPEG2000 code stream
- * @return - > 0 = Length in bytes of encoded JPEG2000 code stream
+ * @return 0 = Length in bytes of encoded JPEG2000 code stream
```
edwardhartnett: You can remove the = as well, so just:

`@return 0  Length in bytes of encoded JPEG2000 code stream
`

But is this correct? If it returns 0, how is that the length of the stream?

edwardhartnett (changes requested):

src/enc_jpeg2000.c
Hang-Lei-NOAA: Will remove that "=".
The return code just show that the code works good. This will be rewords.

edwardhartnett (approved):

edwardhartnett:
Not sure why this is closing the issue.

Next time try exactly:

"Part of #N"

Maybe that will work without closing the issue.

Hang-Lei-NOAA:
Okay, Thanks.

PR #75, opened 2021-07-15
Add doxygen into getdim.f Part of #55

Hang-Lei-NOAA:
Doxygen has been added into getdim.f 
It is valided at
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/getdim_8f.html

### DIFF
--- a/src/getdim.f
+++ b/src/getdim.f
@@ -1,33 +1,24 @@
+!>    @file
+!>    @brief This routine converts grid information from a GRIB2 grid
+!>    to GRIB1 GDS info.
+!>    @author Stephen Gilbert @date 2002-12-11
+!>
+
+!>    This subroutine returns the dimensions and scanning mode of
+!>    a grid definition packed in GRIB2 Grid Definition Section 3 format.
+!>
+!>    @param[in] csec3 Character array that contains the packed GRIB2 GDS.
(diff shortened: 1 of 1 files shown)


src/getdim.f
```diff
@@ -1,33 +1,24 @@
+!>    @file
+!>    @brief This routine converts grid information from a GRIB2 grid
```
GeorgeGayno-NOAA: @Hang-Lei-NOAA I don't see any references to GRIB1 in this routine. Is this description correct?

Hang-Lei-NOAA: This was not. updated

src/getdim.f
```diff
@@ -1,33 +1,24 @@
+!>    @file
+!>    @brief This subroutine returns the info of GRIB2 Grid Definition
+!>    Section 3 format..
```
GeorgeGayno-NOAA: @Hang-Lei-NOAA I am getting this doxygen warning:  _warning: End of list marker found without any preceding list items_

If I remove the extra period on line 3, the warning is eliminated.

Hang-Lei-NOAA: Yes, it is updated. Thanks!

GeorgeGayno-NOAA (approved):
Looks good. Approved.

PR #76, opened 2021-07-15
Add doxygen in getfield.f, Part of #55

Hang-Lei-NOAA:
The doxygen has been added into the file.
It is valiaded at
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/getfield_8f.html#details

### DIFF
--- a/src/getfield.f
+++ b/src/getfield.f
@@ -1,121 +1,99 @@
+!>    @file
+!>    @brief This subroutine obtain grid information and the unpacked data
+!>    for a field from a GRIB2 file.
+!>    @author Stephen Gilbert @date 2000-05-26
+!>
+
+!>    This subroutine returns the Grid Definition, Product Definition,
+!>    Bit-map (if applicable), and the unpacked data for a given data
+!>    field. Since there can be multiple data fields packed into a GRIB2
+!>    message, the calling routine indicates which field is being requested
+!>    with the ifldnum argument.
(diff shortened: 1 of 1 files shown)


src/getfield.f
```diff
@@ -1,121 +1,99 @@
+!>    @file
+!>    @brief This subroutine obtain grid information and the unpacked data
```
edwardhartnett: The brief statement applied to the whole file, not just the first subroutine. (In most of these files, there is only one subroutine, but in this file there is more than one.)

So it should be something like:

!> @brief Contains subroutines to get fields from...

edwardhartnett (changes requested):

src/getfield.f
Hang-Lei-NOAA: You are right.

Hang-Lei-NOAA: The first subroutine is the main one. Following 4 are to unpack data from section 3-6.

GeorgeGayno-NOAA (approved):
Looks good.

src/getfield.f
edwardhartnett: OK, put a summary of the file in the file @brief...

Hang-Lei-NOAA: Okay

edwardhartnett (approved):

PR #78, opened 2021-07-15
Update the doxygen pages Part of #73

### LINKED ISSUE #73: update gh-pages branch with newly built documentation
@Hang-Lei-NOAA do you know how to update the gh-pages branch with newly built documenation?

Normally we only do this when we release a new version, but it's also worth doing from time to time when doing these documentation fixes, because it gets the built documents out there where people can see them.

If you know how to do it, you can do so whenever you like during this documentation campaign (but in normal times, you should only do it when a new version is released).

If you don't know how, ask, and I will dig up the instructions...

Hang-Lei-NOAA:
update for progress, it is shown on
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/index.html


edwardhartnett (approved):
The gh-pages branch is not a protected branch - you can just update it from the command line without a PR.

Hang-Lei-NOAA:
Thanks Edward, will directly commit.

PR #80, opened 2021-07-16
Add doxygen into getgb2.f, Part of #55

Hang-Lei-NOAA:
Part of #55 
Doxygen has been added into getgb2.f. It was shown on:
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/getgb2_8f.html

### DIFF
--- a/src/getgb2.f
+++ b/src/getgb2.f
@@ -210,12 +210,6 @@ C>    - 97 error reading grib file
 C>    - 99 request not found
 C>    - other gf_getfld grib2 unpacker return code
 C>
-C>    Subprograms called:
-C>    - GETIDX get index
-C>    - GETGB2S search index records
-C>    - GETGB2R read and unpack grib record
-C>    - GF_FREE frees memory used by gfld
-C>
 C>    @note specify an index file if feasible to increase speed.
 C>    do not engage the same logical unit from more than one processor.
 C>    Note that derived type gribfield contains pointers to many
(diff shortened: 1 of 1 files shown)


GeorgeGayno-NOAA (changes requested):
Doxygen does not recognize the "%" in your comments. Try adding a backslash before the %

Hang-Lei-NOAA:
@GeorgeGayno-NOAA 
As you suggested, it works! Thanks!

Hang-Lei-NOAA:
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/getgb2_8f.html

src/getgb2.f
```diff
@@ -1,291 +1,235 @@
...
+C>    - gfld\%fld Array of gfld\%ndpts unpacked data points. This element
+C>    is actually a pointer to an array that holds the data.
+C>    @param[out] IRET integer return code
+C>    - 0 all ok
+C>    - 96 error reading index
+C>    - 97 error reading grib file
+C>    - 99 request not found
+C>    - other gf_getfld grib2 unpacker return code
+C>
+C>    Subprograms called:
```
edwardhartnett: You can remove the section "subprograms called" everywhere you see it. Doxygen has a neat feature we can turn on which automatically provides a list of all subprograms called, which we then don't have to maintain. (If we are maintaining it, it inevitably will become wrong over time, whereas doxygen figures it out automatically each time and is always correct).

edwardhartnett (approved):

src/getgb2.f
Hang-Lei-NOAA: Thanks for the knowledge!

edwardhartnett (approved):

GeorgeGayno-NOAA (approved):
Looks good.

PR #102, opened 2021-07-29
Clean up of docs/user_guide.md #81

### LINKED ISSUE #81: Clean up of docs/user_guide.md
The file docs/user_guide.md is the landing page for the doxygen generated documentation. It contains a high-level description of the library.

The current page is the first pass of converting some legacy documentation, and it needs some more work.

Most importantly, all functions/subroutines need to have "()" added after their name, so that they get hyperlined to the proper subprogram documentation. For example:

`To start a new GRIB2 message, call subroutine GRIBCREATE. GRIBCREATE encodes Sections 0 and 1 at the beginning of the message. This routine must be used to create each message.
`

Should become:

`To start a new GRIB2 message, call subroutine GRIBCREATE(). GRIBCREATE() encodes Sections 0 and 1 at the beginning of the message. This routine must be used to create each message.
`

This will cause the subprogram name GRIBCREATE to automatically be linked, so that the user can click on it and jump to the documentation for GRIBCREATE. This needs to happen with all subprogram names in the document.

Further down in the user_guide.md, there is some example code which is not being formatted correctly in doxygen, and that needs to be fixed.

Hang-Lei-NOAA:
issue #81 
This has been add and tested on
https://hang-lei-noaa.github.io/NCEPLIBS/NCEPLIBS-g2/html/files.html

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -3,10 +3,10 @@
 # Introduction
 
 This document briefly describes the routines available for encoding/decoding
-GRIB Edition 2 (GRIB2) messages.  A basic familiarity with GRIB is assumed.
+GRIB Edition 2 (GRIB2) messages. A basic familiarity with GRIB is assumed.
 
 A GRIB Edition 2 message is a machine independent format for storing
-one or more gridded data fields.  Each GRIB2 message consists of the 
+one or more gridded data fields. Each GRIB2 message consists of the 
 following sections:
 
 <pre>
@@ -22,7 +22,7 @@ SECTION 8 - End Section                   }             }            }
 </pre>
 
 Sequences of GRIB sections 2 to 7, 3 to 7, or sections 4 to 7 may be repeated
-within a single GRIB message.  All sections within such repeated sequences
+within a single GRIB message. All sections within such repeated sequences
 must be present and shall appear in the numerical order noted above.
 Unrepeated sections remain in effect until redefined.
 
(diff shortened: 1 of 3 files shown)


docs/user_guide.md
```diff
@@ -36,72 +36,72 @@ a number of different grids, several routines are used to encode a message.
...
 within one or more GRIB2 messages.
 
-To start a new GRIB2 message, call subroutine GRIBCREATE.  GRIBCREATE 
-encodes Sections 0 and 1 at the beginning of the message.  This routine 
+To start a new GRIB2 message, call subroutine GRIBCREATE(). GRIBCREATE()
+encodes Sections 0 and 1 at the beginning of the message. This routine 
 must be used to create each message.
 
-Subroutine ADDLOCAL can be used to add a Local Use Section ( Section 2 ).
+Subroutine ADDLOCAL() can be used to add a Local Use Section (Section 2).
```
GeorgeGayno-NOAA: I see you added () to the subroutine names. However, doxygen did not create links to these routines.

edwardhartnett (approved):

docs/user_guide.md
Hang-Lei-NOAA: Do you know how to set it?
I add *.mod to the files, but it still not treat this file as code.

GeorgeGayno-NOAA: I was looking at other repos for an example, but did not find one. Maybe it is not possible to create links to source code.

edwardhartnett: Try lower-casing the function name...

Hang-Lei-NOAA:
Okay, thanks both.

docs/user_guide.md
Hang-Lei-NOAA: Lower-casing works.

PR #105, opened 2021-08-31
Add `save`to idxlist to improve performance

### LINKED ISSUE #104: g2 performance degradation
There are reports of g2's performance with between versions 3.1 and 3.4.

There have been few actual code changes so I suspect it's this commit which removes the `save` attribute and deallocates/re-allocates an array if it's already already allocated.

https://github.com/NOAA-EMC/NCEPLIBS-g2/commit/c51c8f0cf2beaa574eecf2e2bbbfaa990c2406f4

https://github.com/NOAA-EMC/NCEPLIBS-g2/pull/52

kgerheiser:
Fix #104 

Tested on WCOSS2 and the runtime was brought down from 25s with v3.4.1 to 2s.

@BoiVuong-NOAA will this break #52 where duplicate I/O units caused an error?

I'm pretty sure that was fixed by this line:

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/7e78071c76e23132acf1d5dd75d0250953bd6a25/src/getidx.f#L101

And the `save` had nothing to do with it.

I wonder if I could ask @BijuThomas-NOAA whether this build works for you? I can't find the original reproducer you sent me.

### DIFF
--- a/src/getidx.f
+++ b/src/getidx.f
@@ -58,7 +58,7 @@ C-----------------------------------------------------------------------
          character(len=1),pointer,dimension(:) :: cbuf
       END TYPE GINDEX
      
-      TYPE(GINDEX) :: IDXLIST(10000)
+      TYPE(GINDEX), save :: IDXLIST(10000)
 
       DATA LUX/0/
 C - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


BoiVuong-NOAA (approved):
Should we wait for Thomas to test with your changes ?  If it is working, then we can implement it.

kgerheiser:
~~He tested it and sent the results in that email chain. The runtime went from 25 seconds to 2 seconds.~~

Sorry, misread that and thought you meant Scott. Yes, I want to make sure we didn't break anything else.

BoiVuong-NOAA:
I meant. Mr. Thomas needs to test with open same unit file number.  Did  
he reply back to you ?

PR #127, opened 2021-09-28
require jasper-2.0.25

### LINKED ISSUE #45: need to require a specific minimum jasper version in cmake (and it's version 2.0.25)
We need to require a specific jasper version.

edwardhartnett:
Fixes #45


Hang-Lei-NOAA (approved):

kgerheiser:
One implication I can think of is that WCOSS-Cray only has Jasper 1.900.1.

edwardhartnett:
Well seems like we have to make the min version 1.900.1...

Hang-Lei-NOAA:
2.0.25 is required on other machines to address the error in grib code.

edwardhartnett:
Well why does  WCOSS-Cray not need an updated jasper?

Hang-Lei-NOAA:
Either the system administrator or we just did not install it. It is said
that Cray will retire soon.

edwardhartnett:
Does the Cray use the intel compiler?

kgerheiser:
All of the supercomputers use Intel compilers. Your logic is close, but it's broken with version 2.0.something to 2.0.25 where the bug was present.

edwardhartnett:
@kgerheiser I don't understand your comment. How is the current logic broken?

OK, I get it, you mean I am being too strict on the intel compiler, and some versions before 2.0.25 actually worked. Well, that's OK, I don't usually track compile problems in that much detail. I think just saying 2.0.25 is required is fine.

kgerheiser:
I think requiring 2.0.25 is fine, but then we need 2.0.25 on WCOSS-Cray so it works there.

edwardhartnett:
Can we install 2.0.25 on the WCOSS-Cray?

arunchawla-NOAA:
I do not think we are doing future testing with wcoss-cray. That system is in maintenance mode only. I suggest focus on the newer systems

Hang-Lei-NOAA:
Yes, Arun

edwardhartnett:
OK, if we can move forward without wcoss-cray, we can check the version and ensure it is 2.0.25. Stand by and I will fix this PR...

Issue #26, opened 2020-10-06
Add tests


edwardhartnett:
Our CI testing does literally nothing for this project. So the good news is, it won't be hard to make it better! ;-)

edwardhartnett:
@Hang-Lei-NOAA and @BoiVuong-NOAA are there any GRIB2 files we can use as samples for tests? That is, some actual data files that are hopefully not too large?

BoiVuong-NOAA:
I create a grib2 file which it has 70 records for testing.  The file size is  52MB
The location is /gpfs/dell2/ptmp/Boi.Vuong/output/for_Ed on MARS

edwardhartnett:
How do you create the file? You have some code to do that?

BoiVuong-NOAA:
No, I do not have any code.  I am using WGRIB2 to extract selected fields from GFS grib2 file.
The syntax is following:

$WGRIB2  gfsfile.grib2 | grep -F text_file | $WGRIB2 -i -grib   new_gfsfile.grib2  gfsmaster.grib2
Where:
-gfsfile.grib2   is any gfs grib2 file (0.25,0.5 or 1.0deg)
-text_file        is a file that contains all TMP, HGT, RH,  etc.     Note: The more fields you have, the filesize will be larger file.

edwardhartnett:
OK, I have found a small GRIB2 file here:
https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod/gdas.20210922/00/wave/gridded/

Do we think that NCEPLIBS-g2 will be able to read this file?

My first question is how to feed this to g2_info, which takes a buffer not a filename. Is the idea that I open the file and read the whole file into a buffer before calling g2_info()? Surely that cannot work with a giant GRIB2 file.

So given a giant GRIB2 file, how would I use NCEPLIBS-g2 to learn about it? What function would I call first?

edwardhartnett:
The current state of play: 19%
![image](https://user-images.githubusercontent.com/38856240/141493693-96899666-9be2-4010-9c6c-86cefb3ab6e5.png)

edwardhartnett:
After recent tests from @Hang-Lei-NOAA and @BrianCurtis-NOAA were merged, we are now at 37% coverage. Big improvement - almost double since early November, so great work guys.

But keep it up - there are plenty of tests still to write! What this coverage means is 6 out of 10 lines of this crucial software are completely untested. We can and will do better...

[...]

PR #130, opened 2021-09-29
Adding a test creating a GRIB2 message

### LINKED ISSUE #140: Add address santize to GNU CI builds
We  get a lot of benefit from simply adding:

-fsantize=address

to the FFLAGS. Memory use is checked and great diagnostic message are printed on errors, which pinpoint the problem. Better than valgrind! ;-)

edwardhartnett:
Add a test that creates and then decodes a message.

This test can be further developed in a future PR.

Part of #132
Part of #26
Fixes #140


BoiVuong-NOAA:
The code table PDT 4.0 is at this link: https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_temp4-0.shtml
For more information, NCEP WMO GRIB2 documentation is at this link: https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/

edwardhartnett:
OK, so what is a valid Product Definition Template Number? That is not clear to me...

BoiVuong-NOAA:
The Product Definition Template Number is 0, 8 or 11,.. of PDT in GRIB2 
table.  In GFS, we have Product Definition Template Number 0 or 8.

edwardhartnett:
OK, where is the Product Definition Template Number defined as 0, 8, or 11? Is there a table that contains that info?

Why not -1, 7, or 13?

Where are these numbers listed?

edwardhartnett:
OK, I changed the Product Definition Template Number to 8, the test still crashes. Let's move on to the next parameter:

```
  ! Contains the data values for the specified Product Definition
  ! Template (N=ipdsnum). Each element of this integer array contains
  ! an entry (in the order specified) of Product Defintion Template
  ! 4.N.

```

What does this mean? What does "4.N" mean in the above?

BrianCurtis-NOAA:
When I run this test it gives me 
```
Output:
----------------------------------------------------------
 Testing g2 library encoding.
 gribend: Section 8 can only be added after Section 7.
 gribend: Section            3  was the last found in given GRIB message.
STOP 4
```
Seems like it wants a few more sections. (I couldn't ascertain whether all sections were mandatory)
Try adding 4 and 7 to start and see if it goes through.

edwardhartnett:
OK, I plan on further extending this test, but it has reached a good point to merge to develop and start running in CI.

edwardhartnett:
@Hang-Lei-NOAA and @BrianCurtis-NOAA can you please review? I would like to get this test into play...

BrianCurtis-NOAA (approved):
I think we need to define some outlines for doxygen formatting that will apply to all NCEPLIBS (at least to start) codes. I note for example that you add "Initial" to first author entries in author log, and we've just added tables to the program log format. I think they're all good ideas. Maybe they should be docuented somewhere and applied everywhere. That way new code doxygen can follow it too.

PR #159, opened 2021-10-06
added bacio back

### LINKED ISSUE #154: Function baread() from bacio is called from src/getgb2r.f, but bacio is neither found nor linked...
In subroutine GETGB2R() there are calls to the bacio library:

For example note the calls to BAREAD() around line 206:

```
C  READ AND UNPACK BIT_MAP, IF PRESENT
      IF ( gfld%ibmap.eq.0.OR.gfld%ibmap.eq.254 ) THEN
         ISKIP=LSKIP+SKIP6
         CALL BAREAD(LUGB,ISKIP,4,LREAD,CSIZE)    ! GET LENGTH OF SECTION
         CALL G2_GBYTEC(CSIZE,ILEN,0,32)
         ALLOCATE(CTEMP(ILEN))
         CALL BAREAD(LUGB,ISKIP,ILEN,LREAD,CTEMP)  ! READ IN SECTION
```

This seems to indicate that this subroutine is never used anywhere. If so, we should drop it...

@BoiVuong-NOAA and @Hang-Lei-NOAA do you know anything about this? Was there a past dependency on bacio? Is the subroutine GETGB2R() expected to work by anyone?

edwardhartnett:
Fixes #154 

Adding bacio back into NCEPLIBS-g2.

### DIFF
--- a/.github/workflows/Intel.yml
+++ b/.github/workflows/Intel.yml
@@ -32,6 +32,21 @@ jobs:
         sudo apt-get install intel-oneapi-dev-utilities intel-oneapi-mpi-devel intel-oneapi-openmp intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
         echo "source /opt/intel/oneapi/setvars.sh" >> ~/.bash_profile
       
+    - name: checkout-bacio
+      uses: actions/checkout@v2
+      with:
+        repository: NOAA-EMC/NCEPLIBS-bacio
+        path: bacio
+        ref: develop
+
+    - name: build-bacio
+      run: |
+        cd bacio
+        mkdir build
+        cd build
+        cmake .. -DCMAKE_INSTALL_PREFIX=~/bacio
+        make -j2
+        make install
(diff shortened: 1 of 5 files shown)


Hang-Lei-NOAA (approved):
You choose the safe way.

kgerheiser (changes requested):
Need `find_dependency(bacio)` in the CMake package config

edwardhartnett:
OK. Seems like jasper is also not being found by the CMake package config...

kgerheiser:
Jasper is actually not necessary to add to `find_dependency`. Jasper doesn't use a CMake target (`jasper::jasper`) to define itself, using `${JASPER_LIBRARIES}` instead. When you link directly to libraries (`-l, -L`) like `JASPER_LIBRARIES` does it directly embeds itself in `INTERFACE_LINK_LIBRARIES` instead of needing to search for the target at build time like the package config does.

Point being that the Jasper library is hardcoded in the package config already because it doesn't use a target. Adding it here would just add it twice.

kgerheiser (approved):

PR #171, opened 2021-11-08
improved CI

### LINKED ISSUE #170: Improve code coverage report, check for warnings, upgrade jasper in CI build, and add macos to the CI builds
Improve code coverage report, check for warnings, upgrade jasper in CI build. Also add MacOS.

edwardhartnett:
Fixes #170


edwardhartnett:
Now the code coverage report looks better:

image

edwardhartnett:
This is failing on the mac with this error:

```
[ 92%] Linking Fortran executable test_simpack
196
Undefined symbols for architecture x86_64:
197
  "_mova2i_", referenced from:
198
      _g2_sbytesc_ in libg2_4.a(g2_gbytesc.f.o)
199
      _g2_gbytesc_ in libg2_4.a(g2_gbytesc.f.o)
200
ld: symbol(s) not found for architecture x86_64
```

@kgerheiser any thoughts on why?

Hang-Lei-NOAA:
Could you please summarize the reason of the macos problem and how you solved it? 
As the code review, I saw that you used the system png solved the problem. But need more to learn the reason. 
As I investigated from other sources, it is said that it could be a Mac issue, and need adding -target x86_64-apple-darwin into flags like "CFLAGS = -g -Wall -Wextra -pedantic -Werror -target x86_64-apple-darwin"

edwardhartnett:
I used the same solution I used in g2c, which I found on stackoverflow. Basically, using find_package only searches the standard dirs, and the macos puts libraries somewhere else in the case of png. I'm not a mac expert but anyway I finally got it working.

I think @aerorahul and @kgerheiser can probably come up with better cmake code to accomplish this...

Hang-Lei-NOAA (approved):

PR #183, opened 2021-11-15
adding test for addgrid(), removing stop call in addgrid()

### LINKED ISSUE #177: workflow error in gnu_bacio-2.4.1.yml is preventing it from running
Invalid workflow file : .github/workflows/gnu_bacio-2.4.1.yml#L5
The workflow is not valid. .github/workflows/gnu_bacio-2.4.1.yml (Line: 5, Col: 3): The identifier 'bacio-2.4.1-build' is invalid. IDs may only contain alphanumeric characters, '_', and '-'. IDs must start with a letter or '_' and and must be less than 100 characters.

### LINKED ISSUE #184: addgrid.f calls stop for an error, when it should just return an error code
In addgrid.f we have:

```
!     Check to see if beginning of GRIB message exists.
      do i = 1, 4
          if (cgrib(i) /= grib(i:i)) then
              print *, 'addgrid: GRIB not found in given message.'
              print *, 'addgrid: Call to routine gribcreate required',
     &             ' to initialize GRIB messge.'
 10           format('"', 4A1, '" /= "GRIB"')
              print 10, cgrib(1:4)
              ierr = 1
              stop 1
              return
          endif
      enddo

```

Note the stop call. We cannot call stop within the library. Instead, we return an error code and let the caller decide whether or not to call stop.

edwardhartnett:
Part of #26 
Fixes #184 
Fixes #177

Adding tests for addgrid().

Also removing a stop call in addgrid().

Also fixed broken workflow.

### DIFF
--- a/.github/workflows/gnu_bacio-2.4.1.yml
+++ b/.github/workflows/gnu_bacio-2.4.1.yml
@@ -2,7 +2,7 @@ name: bacio-2.4.1 build
 on: [push, pull_request]
 
 jobs:
-  bacio-2.4.1-build:
+  bacio-2_4_1-build:
     runs-on: ubuntu-20.04
     env:
       FC: gfortran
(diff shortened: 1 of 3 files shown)


edwardhartnett:
Code coverage is ~21%

image

Hang-Lei-NOAA (approved):

PR #200, opened 2021-12-12
Converting code to free-format .F90 files

### LINKED ISSUE #151: Code files should all have .f90 (or .F90) extension instead of .f
All the code files have a .f extension. Yet they use F90 features.

The compilers are mostly OK with this because they just ignore the extension. But it is confusing and also confuses doxygen.

The simplest answer would be to rename each file to a .f90 extension.

@Hang-Lei-NOAA @BrianCurtis-NOAA @kgerheiser any objections to this?

edwardhartnett:
Fixes #151 

Convert all files to free format, and give them .F90 extensions.


Hang-Lei-NOAA:
This is a large amount of work. Great job! I have reviewed 37 files at this point, and will continue tomorrow.

edwardhartnett:
Actually it was not much work. I just renamed the files. I made an emacs macro that converted the F77 style line continuation into F90 line continuation. The whole process took less than two hours.

We will do this on our other F77 codes as well, in g2tmpl, and grib_util...

Hang-Lei-NOAA (approved):

Hang-Lei-NOAA:
How about this: we can distribute all the works of each lib ( code-convert, doxgen, unit tests) to each person. Others can help review.  Then the work would be faster, responsibility is more clear. You don't have to be too busy. For each guy, we are not facing the changes unexpected.

BrianCurtis-NOAA:
> How about this: we can distribute all the works of each lib ( code-convert, doxgen, unit tests) to each person. Others can help review. Then the work would be faster, responsibility is more clear. You don't have to be too busy. For each guy, we are not facing the changes unexpected.

I'd also like to see a mandatory 2 people required to review before a merge. More eyes on code won't hurt.

edwardhartnett:
Let's approach this in an agile fashion.

I will stop assigning tasks to anyone, but I will assign them to the version, for example: https://github.com/NOAA-EMC/NCEPLIBS-g2/projects/2

When you sit down to program, look at the list of open issues for the next release. Pick one, and move it to the "In Progress" column, and assign it to yourself. Then work on it.

(For testing, instead of grabbing the main issue for adding unit tests, write a new issue for just the test you are adding, and label it as part of the main test issue.)

However, don't hang on to an issue - - only what you can do that day or the next. If an issue is too big to do in one day, then break up the issue into issues that can be done in a day or two...

edwardhartnett:
I'm not in favor of increasing the review requirement. I don't believe we are doing especially effective reviews as it is, so I don't think more requirements would help.

Instead, take more time to dig into each review.

However, right now we have a lot of big PRs with very few code changes. Just documentation, formatting, etc. So those are going to be lightly reviewed. That's just the nature of such PRs.

When we are doing actual code changes, we can make the PRs much smaller and it will be much easier to focus on just what is changing. But before that can happen, we've got to execute our main reformatting and documentation changes...

PR #204, opened 2021-12-14
formatting cleanup

### LINKED ISSUE #202: Consistent tabbing for Fortran files
Having consistent indentation is very helpful for understanding code.

I have run all the files (except reduce.F90 and pack_pg.F90) through the following emacs commands:

M-x indent-region 
M-x delete-trailing-whitespace
M-x untabify

edwardhartnett:
On all code files in src, I did the following emacs commands:

```
M-x indent-region
M-x untabify
M-x delete-trailing-whitespace

```
There are no code changes in this PR, only this reformatting.

Fixes #202


src/compack.F90
```diff
@@ -28,416 +28,416 @@
...
+           if (idrstmpl(17).eq.2) then
+              !  pack second original value
+              if (ival2.ge.0) then
+                 call g2_sbytec(cpack,ival2,iofst,nbitsd)
                  iofst=iofst+nbitsd
               else
                  call g2_sbytec(cpack,1,iofst,1)
                  iofst=iofst+1
-                 call g2_sbytec(cpack,iabs(minsd),iofst,nbitsd-1)
+                 call g2_sbytec(cpack,iabs(ival2),iofst,nbitsd-1)
```
BrianCurtis-NOAA, line 185: Why this change?

edwardhartnett: It's only an apparent change because the code looks so similar in two places and is confusing diff. If you look at the line nubmers you see that line 185 is supposed to have ival3 and line 196 minsd, and both do.

BrianCurtis-NOAA: Wow github, not cool. Thx. I noticed the same thing happening later on, and didn't comment in case it was something like this.

src/dec_png.c
```diff
@@ -86,112 +86,112 @@ int SUB_NAME(unsigned char *pngbuf,g2int *width,g2int *height,char *cout)
...
-**     Check if image was grayscale  
+**     Check if image was grayscale
 */
 
 /*
-    if (color != PNG_COLOR_TYPE_GRAY ) {
-       fprintf(stderr,"dec_png: Grayscale image was expected. \n");
-    }
+  if (color != PNG_COLOR_TYPE_GRAY ) {
+  fprintf(stderr,"dec_png: Grayscale image was expected. \n");
```
BrianCurtis-NOAA, line 174: This should be indented.

Hang-Lei-NOAA (approved):

src/compack.F90
Hang-Lei-NOAA: This is the auto system comparison issue.

src/dec_png.c
edwardhartnett: Who are we to question emacs auto-indent function?

In this case it is not indented because it is commented out code. Them's the breaks...

PR #209, opened 2021-12-21
Add test for getpoly.F90

BrianCurtis-NOAA:
Test for getpoly.F90

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -73,3 +73,7 @@ add_test(test_getgb2 test_getgb2)
 add_executable(test_getdim test_getdim.f90)
 target_link_libraries(test_getdim g2_d)
 add_test(test_getdim test_getdim)
+
+add_executable(test_getpoly test_getpoly.f90)
+target_link_libraries(test_getpoly g2_d)
+add_test(test_getpoly test_getpoly)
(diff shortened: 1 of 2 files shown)


Hang-Lei-NOAA (approved):

tests/CMakeLists.txt
```diff
@@ -73,3 +73,7 @@ add_test(test_getgb2 test_getgb2)
 add_executable(test_getdim test_getdim.f90)
 target_link_libraries(test_getdim g2_d)
 add_test(test_getdim test_getdim)
+
+add_executable(test_getpoly test_getpoly.f90)
+target_link_libraries(test_getpoly g2_d)
+add_test(test_getpoly test_getpoly)
```
edwardhartnett, line 79: Github wants a carrage return at the end of the last line for some reason...

tests/test_getpoly.f90
```diff
@@ -0,0 +1,37 @@
+! test getpoly.F90
```
edwardhartnett, line 1: Something like:

This is a test for the NCEPLIBS-g2 project. It tests the getpoly subroutine.

BrianCurtis-NOAA: We can get this in any future PR, no problem.

IMO seems redundant.

PR #210, opened 2021-12-22
adding intmath test

### LINKED ISSUE #201: Recover test code in intmath.F90
There is some code at the bottom of intmath.F90 like this:

```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
c$$$ TEST PROGRAM FOR THIS MODULE
c$$$      program test_intmath
c$$$      use intmath
c$$$      implicit none
c$$$      real(kind=16) :: temp
c$$$      real(kind=16), parameter :: alog2=log(2.0_16)
c$$$      integer(kind=8), parameter  :: &
c$$$     &     one=1,big=Z'7FFFFFFFFFFFFFFF',small=-2000000_8, &
c$$$     &     check=Z'1FFFFFFF'
c$$$      integer(kind=8) :: ival, iret
c$$$      !$OMP PARALLEL DO PRIVATE(ival,temp,iret)
c$$$      do ival=small,big
c$$$ 10      format(Z16,' -- MISMATCH: ',I0,'=>',I0,' (',I0,' = ',F0.10,')')
c$$$ 20      format(Z16,' -- OKAY:     ',I0,'=>',I0,' (',I0,' = ',F0.10,')')
c$$$         if(ival+one<ival) then
c$$$            temp=log(real(max(ival,one),kind=16))/alog2
c$$$         else
c$$$            temp=log(real(max(ival+one,one),kind=16))/alog2
c$$$         endif
c$$$         iret=i1log2(ival)
c$$$         if(iret/=ceiling(temp) .or. ival==0 .or. ival==check) then
c$$$            !$OMP CRITICAL
c$$$            if(iret/=ceiling(temp)) then
c$$$               print 10, ival, ival, iret,ceiling(temp),temp
c$$$            else
c$$$               print 20, ival, ival, iret,ceiling(temp),temp
c$$$            endif
c$$$            !$OMP END CRITICAL
c$$$         endif
c$$$      enddo
c$$$      !$OMP END PARALLEL DO
c$$$      end program test_intmath
```
 Can this be converted into a useful unit test?

edwardhartnett:
Fixes #201 
Part of #26 

Adding some testing for the intmath.F90 module.

### DIFF
--- a/src/intmath.F90
+++ b/src/intmath.F90
@@ -1,22 +1,23 @@
-!>    @file
-!>    @brief Define math functions used by other code.
-!>    @author Stephen Gilbert @date 2000-06-21
-!>
+!> @file
+!> @brief Define math functions used by other code.
+!> @author Stephen Gilbert @date 2000-06-21
 
-!>    This module defines integer math functions used by other programs.
-!>    It include following functions:
-!>    - ilog2 Calculate log(x)/log(2).
-!>     - ilog2_8 for 8 bit integer numbers.
-!>     - ilog2_4 for 4 bit integer numbers.
-!>     - ilog2_2 for 2 bit integer numbers.
-!>     - ilog2_1 for 1 bit integer numbers.
-!>    - i1log2 Calculate log(x+1)/log(2) unless x=maxint, in which case log(x)/log(2).
-!>     - i1log2_8 for 8 bit integer numbers.
-!>     - i1log2_4 for 4 bit integer numbers.
-!>     - i1log2_2 for 2 bit integer numbers.
-!>     - i1log2_1 for 1 bit integer numbers.
+!> This module defines integer math functions used by compack,
+!> simpack, and misspack.
 !>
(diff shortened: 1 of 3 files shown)


edwardhartnett:
OK this takes testing of intmath.f90 to 80%, which is good. And so this should be merged.

But there are some things in this code I don't understand. Maybe @Hang-Lei-NOAA or @BrianCurtis-NOAA might have a better idea of the math? Here's a section of code I don't get. What is going on with the untested code? Why compare the value to Z'F0'?

image

BrianCurtis-NOAA (approved):

Hang-Lei-NOAA:
I am not sure why they compare with 'F0'. But just think and guess that they would like to selects the smallest positive actual field width ('F0') to avoid crash out.  To understand why this test is not covered, I may need to better understand what you just explain it to me on this.

PR #214, opened 2022-01-04
Add implicit none to addfield

BrianCurtis-NOAA:
Make adjustments to the code to include 'implicit none'.

### DIFF
--- a/src/addfield.F90
+++ b/src/addfield.F90
@@ -82,6 +82,9 @@ subroutine addfield(cgrib, lcgrib, ipdsnum, ipdstmpl, ipdstmplen, &
      idrstmplen, fld, ngrdpts, ibmap, bmap, ierr)
   use pdstemplates
   use drstemplates
+
+  implicit none
+
   logical :: match
   character(len = 1), intent(inout) :: cgrib(lcgrib)
   integer, intent(in) :: ipdsnum, ipdstmpl(*)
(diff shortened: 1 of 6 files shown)


src/addfield.F90
```diff
@@ -269,12 +281,12 @@ subroutine addfield(cgrib, lcgrib, ipdsnum, ipdstmpl, ipdstmplen, &
   if (ibmap .eq. 0 .or. ibmap .eq. 254) then
      allocate(pfld(max(2, ngrdpts)))
      ndpts = 0;
-     do jj = 1, ngrdpts
```
edwardhartnett, line 272: why this change?

BrianCurtis-NOAA: `jj` was being used as that iterator and was being changed by the outputs from the call here:
https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/da8a6c538afb6b4a38e778b8513220d2f11f7010/src/addfield.F90#L313-L315

I didn't want to risk any problems by having two of the same variable name used two different ways.

edwardhartnett: Are we sure this does not change code behavior? (Even if that behavior is a bug?)

BrianCurtis-NOAA: the only use of `jj` is in the three lines i showed previous. The new iterator `ii` is not used elsewhere in the code. In this case the code was fine, there was no cause for concern about reusing `jj` like it was. It's just not smart to have a variable used in two different ways like that.

edwardhartnett (approved):

PR #215, opened 2022-01-11
Add test for putgb2

BrianCurtis-NOAA:
WIP

Working out some bugs in how fld is getting packed from creategrib.f90 and how getgb2 is getting from the file.

Part of #26.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -86,3 +86,6 @@ add_executable(test_intmath test_intmath.f90)
 target_link_libraries(test_intmath g2_d)
 add_test(test_intmath test_intmath)
 
+add_executable(test_putgb2 test_putgb2.f90)
+target_link_libraries(test_putgb2 g2_d)
+add_test(test_putgb2 test_putgb2)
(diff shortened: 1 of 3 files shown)


edwardhartnett:
How's this coming?

BrianCurtis-NOAA:
I'd have to bring it back up (moved to getting specpack unit test to work). The values either coming in from the getgb2 or something with putgb2 wasn't matching. The next step on this is to go line by line. Feel free to try it out. I'll look at both putgb2 and specpack unit tests more this afternoon.

I may need to push what I was testing with, i'll check that out soon.

edwardhartnett:
OK, I took a look at this and I think the problem is a bug in putgb2.F90, where we have this code:

```
  ! Allocate array for grib2 field.
  lcgrib = gfld%ngrdpts * 4

```
And then later:

`     call gribcreate(cgrib, lcgrib, listsec0, gfld%idsect, ierr)
`

Well lcgrib is supposed to be enough to hold the entire message, not just the data portion.

How did this ever work? For large data sizes, with compression, this would give a large number, larger than actually needed to store the data. So this code would seem to work for large grid sizes (such are actually used) and break for toy-sized ones such as we are using in the test.

I will continue to investigate...

edwardhartnett:
I have put this test up in #248 . Turns out there was a memory problem in the code, which this test found, so great work @BrianCurtis-NOAA , that's exactly why we write tests. ;-)

I will close this PR.

PR #227, opened 2022-02-08
Unit test for getg2i

BrianCurtis-NOAA:
Initial add for test_getg2i
* Issue with the file i'm trying to read. I keep getting an iret=4 on the getg2i function call meaning it can't read the index file header.


BrianCurtis-NOAA:
@Hang-Lei-NOAA I am guessing you may know the .idx file well. Do you know if the header is OK for it? I took that file from the g2c tests.

edwardhartnett:
Did you copy the index file into the binary directory in CMake?

Take a look at CMakeLists.txt file in the tests directory, you will see this function defined:

```
# Copy test data file to the build directory.
function(copy_test_data name)
  file(COPY "${CMAKE_SOURCE_DIR}/tests/${name}"
    DESTINATION ${CMAKE_BINARY_DIR}/tests
    FILE_PERMISSIONS OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ)
endfunction()

```

Later in the file is an example of using it:

```
#copy test data files
copy_test_data(testdata_g2grids)

```

If you want to use a data file in testing, you need to add it to the tests directory, and then add the cmake command to copy the file to the binary directories...

Hang-Lei-NOAA:
I try it once, it seems not the issue of copy files. More likely a BACIO link issue.

BrianCurtis-NOAA:
> Did you copy the index file into the binary directory in CMake?
> 
> Take a look at CMakeLists.txt file in the tests directory, you will see this function defined:
> 
> ```
> # Copy test data file to the build directory.
> function(copy_test_data name)
>   file(COPY "${CMAKE_SOURCE_DIR}/tests/${name}"
>     DESTINATION ${CMAKE_BINARY_DIR}/tests
>     FILE_PERMISSIONS OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ)
> endfunction()
> ```
> 
> Later in the file is an example of using it:
> 
> ```
> #copy test data files
> copy_test_data(testdata_g2grids)
> ```
> 
> If you want to use a data file in testing, you need to add it to the tests directory, and then add the cmake command to copy the file to the binary directories...

I did, here. That's what you meant, correct?

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/83d1156b21e811bc8558106dc78031c525b3a089/tests/CMakeLists.txt#L111-L116

edwardhartnett:
OK, I am taking a look at this, and I don't understand how it even compiles (but it does).

Where I write a test program that is simplfied and similar it fails to compile:

```
program test
  implicit none
  character(len=1), pointer, dimension(:) :: cbuf(:)
  print *, 'howdy!'
  call t1(cbuf)
end program test
subroutine t1(cbuf)
  implicit none
  character(len=1),pointer,dimension(:) :: cbuf
  print *, 'h1!'
end subroutine t1
```

This fails to compile:

```
gfortran -o test test.F90 && ./test
test.F90:5:9:

    5 |   call t1(cbuf)
      |         1
Error: Explicit interface required for ‘t1’ at (1): pointer argument 
(base) ed@koko:~$ 
```

@BrianCurtis-NOAA and @Hang-Lei-NOAA , you guys know Fortran better than I do. Can you explain?

BrianCurtis-NOAA:
@edwardhartnett do you need to add `external t1` in your `program` section?

kgerheiser:
@edwardhartnett it's because it's not in a `module` or a `contains`. Passing pointers like that requires an explicit interface, and just listing subprograms next to each doesn't do that unless it's in a module.

So, if you compile like:


```
program test
  implicit none
  character(len=1), pointer, dimension(:) :: cbuf(:)
  print *, 'howdy!'
  call t1(cbuf)

contains

subroutine t1(cbuf)
  implicit none
  character(len=1),pointer,dimension(:) :: cbuf
  print *, 'h1!'
end subroutine t1

end program test

```

It will work.

kgerheiser:
Also, I don't think there's any reason for that to be a pointer. Why not `allocatable`? Or does g2 require passing a pointer?

edwardhartnett:
@kgerheiser I understand what you are saying, but then why does the test code compile, when it uses neither and interface nor a contains?

I doesn't work, it segfaults when cbuf is used in the subroutine. But it compiles.

When I try to do the same in my little example code, I get a compile error. When I look at getg2ir.f90, I see that an interface is used for cbuf...

I don't know why it's a pointer instead of allocatable, but it is passed as an inout parameter, so the buffer does need to be passed to the calling program. Would that work if it were just allocatable?

edwardhartnett:
@BrianCurtis-NOAA when I add external t1 to my little test code I get:

```
gfortran -o test test.F90
test.F90:8:13:

    8 |   external t1
      |             1
Error: Explicit interface required for ‘t1’ at (1): pointer argument
(base) ed@koko:~$ 
```

But your test does not have an external statement. I'm trying to get a smaller test program that has the same problems as yours, but I can't even get it to compile!

edwardhartnett:
I'm going to take this PR down. I am attempting to add this test in #256

PR #238, opened 2022-05-16
Fix building g2 library when it's added as a subdirectory of another project

### LINKED ISSUE #239: Build fails when g2 project is used as a subdirectory in another project
When this project is added as a subdirectory in another project, for example:
```cmake
cmake_minimum_required(VERSION 3.15)

project(read_g2 LANGUAGES Fortran)

add_subdirectory(g2)

add_executable(read_g2 read_g2.f90)

target_link_libraries(read_g2 g2_4)
```
build fails.

Fixed in #238

DusanJovic-NOAA:
* At several places CMAKE_SOURCE_DIR is used instead of CMAKE_CURRENT_SOURCE_DIR, and corresponding CMAKE_BINARY_DIR instead of CMAKE_CURRENT_BINARY_DIR.
* Package configuration is moved from src to top-level CMakeLists.txt
* Fix compiler warning in src/pack_gp.F90
* Add two cmake configuration options (BUILD_4 and BUILD_D)

Fixes #239


kgerheiser (approved):
Looks good to me

PR #242, opened 2022-06-01
removed gdt2gds.F90 from build

### LINKED ISSUE #229: Add w3emc dependency in support of gdt2gds.F90 which is used by grib_util
While writing a unit test for the gdt2gds.F90 code, I noticed that it uses w3emc calls for https://github.com/NOAA-EMC/NCEPLIBS-w3emc/blob/develop/src/w3fi71.f and https://github.com/NOAA-EMC/NCEPLIBS-w3emc/blob/develop/src/r63w72.f at the following two locations:

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/9a52944bb58f77cd6bb88252f00e41f8c98af552/src/gdt2gds.F90#L344

and

https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/9a52944bb58f77cd6bb88252f00e41f8c98af552/src/gdt2gds.F90#L364

There is currently no w3emc dependency inside g2. Also I would assume nobody is using that code considering the missing dependency has been around for a long time.

We should remove it.

edwardhartnett:
Fixes #229


kgerheiser:
Not sure about this without at least a cursory grep of several applications that use g2. Just because there's no explicit CMake dependency doesn't mean this isn't used. Many applications link to w3emc/bacio/w3nco already so the dependency would be satisfied.

edwardhartnett:
Yes, but there are reportedly very few users of GRIB1 left. So I am thinking we take this out, and see if anything breaks.

GeorgeGayno-NOAA:
> Not sure about this without at least a cursory grep of several applications that use g2. Just because there's no explicit CMake dependency doesn't mean this isn't used. Many applications link to w3emc/bacio/w3nco already so the dependency would be satisfied.

Check copygb and copygb2.

kgerheiser:
Yeah, they both use that function.

edwardhartnett:
OK, crap. Never mind!

BrianCurtis-NOAA:
@kgerheiser So if we have a call in gdt2gds that uses w3emc functions but doesn't add that dep in CMake, would we not be doing things wrong in some way? Should we add a dep on w3emc since g2 hosts that file? Should we move that file somewhere else?

kgerheiser:
Yes, there should be a w3emc dependency. It all works because there are no interfaces and we use static libraries.

Issue #191, opened 2021-11-17
implicit none should be used in every subroutine


edwardhartnett:
implicit none should be used in every subroutine and all variables should be explicitly declared.

edwardhartnett:
I will close this issue and if I find anywhere I missed, I'll address it.

Issue #186, opened 2021-11-16
make code more readable by adding spaces


edwardhartnett:
We have code like this:

```
      if ( ibmap.eq.0 .OR. ibmap.eq.254 ) then
         allocate(pfld(max(2,ngrdpts)))
         ndpts=0;
         do jj=1,ngrdpts
             intbmap(jj)=0
             if ( bmap(jj) ) then
                intbmap(jj)=1
                ndpts=ndpts+1
                pfld(ndpts)=fld(jj);
             endif
         enddo

```

Spacesareimportantfortheeyetobeabletoeasilyseparatesymbols.It'spossibletoreadtextwithoutspaces,butit'sharderandtakesmoreeffort.Wewantourcodetobereadable.

So we should use spaces in the normal manner.

Issue #132, opened 2021-09-30
Doxygen cleanup and improvements still needed...


edwardhartnett:
Although we have done a good job converting the legacy docs to doxygen, there is still plenty of improvement possible.

Some possible improvements:
* Further formatting cleanup.
* Adding links.
* Expanding acronyms.
* Adding missing punctuation.

While adding tests, we will be pouring over the documentation. That's a good time to improve it.

edwardhartnett:
Closing this issue before release 3.5.0. Many good changes have been made to documentation in this release...

PR #248, opened 2022-07-15
Fix memory problem in putgb2.F90, and add test program

### LINKED ISSUE #251: not calculating the correct size of a GRIB2 message in puggb2.F90
in putgb2.F90, where we have this code:

```
  ! Allocate array for grib2 field.
  lcgrib = gfld%ngrdpts * 4

```
And then later:

`call gribcreate(cgrib, lcgrib, listsec0, gfld%idsect, ierr)
`

Well lcgrib is supposed to be enough to hold the entire message, not just the data portion.

How did this ever work? For large data sizes, with compression, this would give a large number, larger than actually needed to store the data. So this code would seem to work for large grid sizes (such are actually used) and break for toy-sized ones such as we are using in the test.

edwardhartnett:
Fixing memory problem in putgb2.F90. Also added implicit none and had to add a few new integer definitions.

Adding a test that Hang developed.

Also some code cleanup.

Fixes #251 
Part of #191 
Part of #186 
Part of #132
Part of #26

### DIFF
--- a/.github/workflows/gcc_docs_test-coverage.yml
+++ b/.github/workflows/gcc_docs_test-coverage.yml
@@ -82,7 +82,7 @@ jobs:
         mkdir build
         doxygen --version
         cd build
-        cmake .. -DJasper_ROOT=~/Jasper -DCMAKE_PREFIX_PATH="~/bacio;~/w3emc" -DENABLE_DOCS=Yes -DCMAKE_Fortran_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fsanitize=address" -DCMAKE_C_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fsanitize=address"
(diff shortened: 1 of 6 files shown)


edwardhartnett:
Guys, this is what I added:

```
  ! Figure out the maximum length of the GRIB2 message.
  lcgrib = 16 + 21 + 4 ! Sections 0, 1, and 8.
  ! Check for Section 2.
  if (associated(gfld%local) .AND. gfld%locallen .gt. 0) then
     lcgrib = lcgrib + gfld%locallen * 4
  endif
  ! Maximum size for Sections 3, 4, and 5 < 512 each.
  lcgrib = lcgrib + 512 + 512 + 512
  ! Is there a section 6?
  if (gfld%ibmap .eq. 0) then
     lcgrib = lcgrib + gfld%ngrdpts
  endif
  ! Section 7 holds the data.
  lcgrib = lcgrib + gfld%ngrdpts * 4

  ! Allocate array for grib2 field.
  allocate(cgrib(lcgrib), stat = is)
```

BrianCurtis-NOAA (approved):

Issue #270, opened 2022-07-26
Remove unneeded return statements at the end of subprograms


edwardhartnett:
We have lots of code like this:

```
    return
end subroutine getdim

```

This was required in FORTRAN 66, but has not been required since F77. ;-)

The end subroutine will return from the subroutine. There is no need to put a "return" just before it. Let's remove this historical clutter.

PR #271, opened 2022-07-26
Adding test for getgb2p.F90, also now nullify pointer before use

### LINKED ISSUE #272: nullify not called on pointer before use in getgb2p.F90
The f2003, and later, standards say for
ASSOCIATED (POINTER [,TARGET])

"POINTER ... Its pointer association status shall not be undefined."

So when we use a pointer, and don't nullify it, and then call a line like this:

     if (.not. associated(gribm)) allocate(gribm(leng))

We have a problem. The pointer variable gribm must be nullified before this statement is encountered, or segfault results.

Apparently this is something that only recently started being enforced in GNU compilers.

edwardhartnett:
Part of #26 
Part of #270 
Part of #191 
Part of #132
Fixes #272 

More tests, more doxygen improvements, more code cleanup.

Also now call nullify() on pointer before use in getgb2p.F90.


edwardhartnett:
@Hang-Lei-NOAA this is ready to merge - please review.

Hang-Lei-NOAA (approved):

PR #273, opened 2022-07-27
removed unneeded return statements at the end of subprograms

### LINKED ISSUE #270: Remove unneeded return statements at the end of subprograms

edwardhartnett:
Fixes #270 

Every time we include a line of unnecessary code in our release, an angel kicks a kitten right in the face. Really hard too!

So let's eliminate whatever code we don't need.


Hang-Lei-NOAA:
For me it is a little bit confused, Does the return statement in the end of each subroutine can be removed if there is no value to return?  Since the 'end subroutine' is followed. 
In getfield .f90, only the 'return' in the sixth subroutine was removed. Others are still there.
Could you please further explain it?

edwardhartnett:
Yes, return can be removed in subroutines if it comes just before the end of the subroutine. In other words, the end of the subroutine does the return. Note that in Fortran 66, this return was required. ;-)

I will fix getfield.F90...

edwardhartnett:
@Hang-Lei-NOAA please review...

Hang-Lei-NOAA (approved):

PR #279, opened 2022-07-30
renamed tests to F90, added some forgotten tests, whitespace cleanup of F90 files

### LINKED ISSUE #276: change test names from f90 to F90
Well we changed the source files to all use F90 instead of f90, but we neglected the tests. So that's confusing.

Let's also change the tests and use F90 everywhere.

edwardhartnett:
Fixes #276 
Part of #26 
Part of #186

Rename tests from f90 to F90.

Also added a few tests that somehow had been missed in the CMake file.


edwardhartnett:
More testing of pdstemplates is possible, but I'll circle around to that in another PR...

This one is ready for review.

Hang-Lei-NOAA (approved):
My reviews see that except three new test codes, a few adding several definitions/initializations, Major codes in these 73files are format adjustment.

Issue #181, opened 2021-11-14
Fix all code warnings and enforce warning check with CI


edwardhartnett:
With the -Werror flag, we can forever stop lazy programmers from checking in code with warnings!

But first we must clear up all the warnings in the code. This needs to be done after we have good testing in place.

edwardhartnett:
I have fixed many warnings, but some are built into library functioning. For example see this code from test_gbytec:

```
  ! For this test to pass the -fallow-argument-mismatch flag must be
  ! used, because I am passing in a real array instead of an int array
  ! for the in parameter. This is how g2_sbytesc() is called in
  ! addfield.F90.
  print *, 'Testing g2_sbytesc() with a real array (size 1) instead of an int array...'
  iskip = 0
  nbits = 32
  nskip = 0
  num = 1
  r_in(1) = 1 
  call g2_sbytesc(out4, r_in, iskip, nbits, nskip, num)

```

PR #326, opened 2022-11-21
fixed warnings

edwardhartnett:
Part of #181 

Fixed some warnings.


edwardhartnett:
@Hang-Lei-NOAA or @AlexanderRichert-NOAA please review...

Hang-Lei-NOAA (approved):
Is it better to assign the value after define the variables?

edwardhartnett:
@Hang-Lei-NOAA I think it's fine to set the variables at initialization just to prevent warnings.

In all cases here, the variables that are generating the warning are only set, and only used, under some circumstances. Under all others, the varible never gets defined or used. So this generates a warning, but setting them to 0 when defined will quiet the warning and have no effect on the code.

(And this is what the intel compiler does to ints by default anyway.)

PR #327, opened 2022-11-21
Expose the g2c API in Fortran

edwardhartnett:
The g2c library contains some new g2c_* functions. I need to use them in grb2index testing in NCEPLIBS-grib_util. For that to work well, it would be best to expose these functions in the g2 library.

In the g2 library they will all have the prefix: "g2cf".

So g2cf_open() will open a GRIB2 file, and g2cf_close() will close it, etc.

Part of #328


edwardhartnett:
OK, when this is merged to develop, I can then use these functions to resolve the grib_util testing problems with grb2index.

edwardhartnett:
This will not pass CI until this g2c PR is merged: https://github.com/NOAA-EMC/NCEPLIBS-g2c/pull/384

AlexanderRichert-NOAA (approved):
Test on local machine looks good

PR #332, opened 2022-12-01
CI improvements

edwardhartnett:
Part of #331.

Improving CI system.

When I turn on the Linux_versions.yml workflow, it breaks due to #333 . I will circle around to that in my next PR.


Hang-Lei-NOAA (commented):
I am thinking that changing the CI target from the develop branch to the target release version may not be good. Some developers may forget to update the workflow when they contribute to the development.

edwardhartnett:
@Hang-Lei-NOAA there is a developer.yml which tests the develop branch of all NCEPLIBS dependencies.

There is a Linux_versions.yml that tests all combinations of acceptable released versions of dependencies.

So changes will be tested against both develop branches, and all recent releases, of all NCEPLIBS dependencies.

As we do releases, we have to add them to the testing of packages that use them. For example, I just did a w3emc release, so once I get the CI systems working well, I will make sure that every Linux_version.yml includes the new w3emc version.

edwardhartnett:
Also @Hang-Lei-NOAA please approve this PR so I can move on to the next phase...

Hang-Lei-NOAA (approved):

Issue #333, opened 2022-12-01
segfaults in some tests under some configurations of CI


edwardhartnett:
As part of the CI improvements, I find:

```
48/58 Test #48: test_putgb2_d ....................***Exception: SegFault  0.11 sec
 gfld%ngrdpts from getgb2 is:            4
 gfld%fld from getgb2 is:    1.1000000000000001        1.3000000000000000        1.3000000000000000        1.3000000000000000     
 Testing open/write/close of GRIB2 file

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7fafe92a6ad0 in ???
#1  0x7fafe92a5c35 in ???
#2  0x7fafe8e9751f in ???
#3  0x55946461fdb8 in putgb2_
	at /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/src/putgb2.F90:141
#4  0x5594646148ad in test_putgb2
	at /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_putgb2.F90:86
#5  0x55[946](https://github.com/NOAA-EMC/NCEPLIBS-g2/actions/runs/3593961072/jobs/6051602544#step:15:947)461429e in main
	at /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_putgb2.F90:5
The following tests FAILED:
1160
	 19 - test_putgb2_4 (SEGFAULT)
1161
	 48 - test_putgb2_d (SEGFAULT)
```

Issue #329, opened 2022-11-29
more testing


edwardhartnett:
More testing.

edwardhartnett:
Many tests have been added. I will close this issue and new tests will be added under their own issues.

PR #336, opened 2022-12-02
test code cleanup in preparation for some debugging of memory issues

edwardhartnett:
Cleanup of some test code.

There is something weird going on in these test code, it causes a memory error under the correct conditions, see #333 .

In this PR I clean up the code but make no changes. After this is merged I will take another pass and see if I can find the memory problems. Meanwhile I am adding a test in the g2c library based on the same GRIB2 file, to make sure the file is OK.

Part of #329.

### DIFF
--- a/tests/test_putgb2.F90
+++ b/tests/test_putgb2.F90
@@ -1,6 +1,7 @@
 ! This is a test program in the NCEPLIBS-g2 project.
 !
-! This program tests the putgb2() subroutine.
+! This program tests the putgb2() subroutine. This program depends on
+! the testgrib.grb2 file alread existing.
 !
 ! Brian Curtis 11/26/2021
 ! Ed Hartnett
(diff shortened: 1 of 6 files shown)


Hang-Lei-NOAA (approved):

tests/test_putgb2.F90
```diff
@@ -22,79 +26,41 @@ program test_putgb2
...
-  end if
+  if (iret .ne. 0) stop 107
 
+  ! Close file.
   call baclose(2, iret)
-  if (iret .ne. 0) then
-     print *, 'baclose failed with iret value: ', iret
-     stop 8
-  end if
+  if (iret .ne. 0) stop 150
```
AlexanderRichert-NOAA, line 59: For my own edification, what is the significance of the "stop" arguments? Are those error codes that are documented somewhere?

edwardhartnett: No significance whatever. They are not error codes.

They could all be the same number in the test but then when it fails, you don't know where. So if you give each stop a different number, you can easily see where the failure is and fire up the debugger.

Test code is not user code, we don't have to give decent error messages or anything. (And that's why I took out these error messages in the test code.) The user should never see a test code failure, so it's only for us developers. Knowing where the stop happened is usually enough to know what is going on...

edwardhartnett:
The Macs are slow today!

I will merge this when they complete...

PR #338, opened 2022-12-02
fixing test, adding logging capability

### LINKED ISSUE #337: test_getgb2.F90 has some commented out code which indicate test problems
In tests/test_getgb2.F90 we have:

```
  ! NOT WORKING
  ! do i = 1, 5
  !   print *, gfld%idrtmpl(i), idrstmpl(i)
  !   ! if (gfld%idrtmpl(i) .ne. idrstmpl(i)) stop 200
  ! end do
  if (gfld%unpacked .neqv. .false.) stop 201
  if (gfld%ibmap .ne. 255) stop 203
  !REMAINING ITEMS NOT WORKING RIGHT
  ! if (gfld%bmap(1) .neqv. .false.) stop 204
  ! do i = 1, 4
  !     print *, gfld%fld(i), fld(i)
  ! !   ! if (gfld%fld(i) .ne. fld(i)) stop 205
  ! end do

```

edwardhartnett:
Fixes #337 
Fxies #341 
Part of #331 

A couple of minor fixes to the test code to remove some commented-out code.

Also add LOGGING capability.

Also improved CI.


AlexanderRichert-NOAA (commented):
test_getgb2_4 is failing on my personal machine with intel compiler.

```
Test project /n/g2/NCEPLIBS-g2/build/tests
    Start 11: test_getgb2_4
1/1 Test #11: test_getgb2_4 ....................***Failed    0.02 sec
 Testing open/read/close of GRIB2 file created with creategrib.f90..
 testing getgb2()..
 1.10000002384186 1.20000004768372 1.29999995231628
 1.39999997615814
 -1.0842022E-19 1.800000 2.000000 2.000000
 -1.0842022E-19 1.10000002384186 1.10000002384186
205
```

edwardhartnett:
There is something weird going on with this code, either the test code (most likely) or the library code has a bug. ;-)

What we're seeing in this test is the data read from the file give the wrong numbers. I can now read this GRIB2 file with the C library (independent code) and also the degrib2 utility (based on g2 code). And the file contains the correct values, See https://github.com/NOAA-EMC/NCEPLIBS-g2c/pull/392.

So I am circling around this bug and will squash it soon!

Hang-Lei-NOAA:
If the value, type in line 90 doesn't match, could fail in this.

Hang-Lei-NOAA (approved):

edwardhartnett:
OK, the problem was pretty simple. The fld array in creategrib.F90 was always being declared with kind=8, even on kind-4 builds. So that was messing up the data.

In order to fix I had to make creategrib a plain subroutine instead of a module, so it could be compiled with the tests and get either KIND_4 or KIND_D as a pre-processor define. Then I changed the fld declaration to this:

```
#ifdef KIND_4  
  real(4) :: fld(ngrdpts) = (/ 1.1, 1.2, 1.3, 1.4 /)
#else
  real(8) :: fld(ngrdpts) = (/ 1.1, 1.2, 1.3, 1.4 /)
#endif

```

edwardhartnett:
I think one of the problems is that there was code to create a test file in test_getgb2.F90, and that test file was also used in test_putgb2.F90. But since the tests are run in parallel, that doesn't always work.

By combing the two tests, I eliminate this problem.

edwardhartnett:
OK, I have now turned on all the testing and everything passes...

edwardhartnett:
OK, we've hit the daily MacOS slowdown. ;-) I will merge when the MacOS runs complete.

PR #343, opened 2022-12-22
porting changes from develop branch to v3.4.x branch, to prepare for a 3.4.6 release

### LINKED ISSUE #189: gf_getfld() leaks memory on error
In gf_getfld() we have:
```
!         If found Section 1, decode elements in Identification Section.
          if (isecnum.eq.1) then
              iofst = iofst-40    ! reset offset to beginning of section
              call gf_unpack1(cgrib, lcgrib, iofst, gfld%idsect,
     &             gfld%idsectlen, jerr)
              if (jerr.ne.0) then
                  ierr = 15
                  return
              endif
          endif

!         If found Section 2, Grab local section. Save in case this is
!         the latest one before the requested field.
          if (isecnum.eq.2) then
              iofst = iofst-40    ! reset offset to beginning of section
              if (associated(gfld%local)) deallocate(gfld%local)
              call gf_unpack2(cgrib, lcgrib, iofst, gfld%locallen,
     &             gfld%local, jerr)
              if (jerr.ne.0) then
                  ierr = 16
                  return
              endif
          endif

```

The gf_unpack1() function allocates an array in gfld. If the function later exits, this memory is not freed and causes a leak.

The solution is to call gf_free() on gfld before an error exit.

There are a bunch of currently commented-out tests that demonstrate this bug in test_getfield.f90.

### LINKED ISSUE #246: Missing find_dependency(w3emc)
The CMake package config should contain `find_dependency(w3emc)`

### LINKED ISSUE #250: turn on address santizer in CI
turn on address santizer in CI

### LINKED ISSUE #303: all tests should run with both _d and _4 version of the library
all tests should run with both _d and _4 version of the library

### LINKED ISSUE #312: parameter TWLDT has wrong parameter number, 200 should be 209
From @BoiVuong-NOAA in #309 

> This line data paramlist(1002) /gribparam(255, 255, 10, 3, 200, 'TWLDT')/
> should be
> data paramlist(1002) /gribparam(255, 255, 10, 3, 209, 'TWLDT')/
> 

So should this be fixed? If not, why not?

### LINKED ISSUE #345: addfield incorrectly handles coordlist size for "d" builds
addfield incorrectly handles coordlist size for "d" builds.

mkieee() always wants a 4-byte integer coord list, but when built with "d" gets an 8-byte one.

### LINKED ISSUE #346: Code depends on bacio and w3emc, but they are not required by cmake
They should be.

edwardhartnett:
Part of #344 
Fixes #345 
Fixes #346
Fixes #312
Fixes #303
Fixes #250
Fixes #246
Fixes #189

Note that this PR is not to develop, but to the new  branch "v3.4.x". This branch is based on the 3.4.5 release.

This PR brings some changes from develop on to the v3.4.x branch, so we can do a 3.4.6 release.


Hang-Lei-NOAA (approved):
Should a code merge from the develop branch to a specific working branch need to go through the review procedure?  The final changes will be brought back to develop branch after going through the review check.

edwardhartnett:
@Hang-Lei-NOAA in this case the branch will be used to do an intermediate release, 3.4.6.

This allows me to release some bug fixes without releasing all the changes of 3.5.0. In 3.5.0 we add a dependency on g2c.

To support this I have created the 3.4.x branch, which will be treated like the develop branch (i.e. with GitHuib protection rules, which I have set up.)

If we need to do a 3.4.7 release, we will do it from the 3.4.x branch, instead of the develop banch. When we are ready for the 3.5.0 release from the develop branch we can start to ignore the v3.4.x branch and eventually delete it.

PR #353, opened 2023-01-29
removed ill-advised README.md excemptions from testing

edwardhartnett:
Part of #331 

More CI cleanup.

### DIFF
--- a/.github/workflows/Intel.yml
+++ b/.github/workflows/Intel.yml
@@ -1,15 +1,16 @@
+# This is a GitHub actions workflow for the NCEPLIBS-g2 project.
+#
+# This test runs an Intel compiler build.
+#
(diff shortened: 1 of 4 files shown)


Hang-Lei-NOAA (approved):

AlexanderRichert-NOAA (approved):
Fine by me. @edwardhartnett I'm curious what makes these ill advised? Surely README.md changes aren't going to break the code?

edwardhartnett:
What makes them ill-advised is that it does not cause the "required" tests to not be required. It just doesn't run them. So there is no way to merge such a PR. ;-)

So, it's a good idea if you are not requiring tests to pass before merging. But that's a pretty important thing to require, IMO.

Issue #318, opened 2022-10-15
documentation improvements


edwardhartnett:
documentation improvements

edwardhartnett:
There have been many documentation improvements. I will close this issue and do further documentation work in new issues.

PR #369, opened 2023-05-10
whitespace cleanup

edhartnett:
Part of #318

### DIFF
--- a/src/enc_png.c
+++ b/src/enc_png.c
@@ -67,8 +67,8 @@ void user_write_data(png_structp png_ptr,png_bytep data, png_uint_32 length)
  * @author Stephen Gilbert
 */
 
-void user_flush_data(png_structp png_ptr)
-
+void 
+user_flush_data(png_structp png_ptr)
 {
    /* int *do_nothing=NULL; */
 }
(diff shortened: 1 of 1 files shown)


src/enc_png.c
```diff
@@ -85,100 +85,76 @@ void user_flush_data(png_structp png_ptr)
...
-*/
 
-    /*bytes=bit_depth/8;*/
-    bytes=*nbits/8;
-    row_pointers=malloc((*height)*sizeof(png_bytep));
-    for (j=0;j<*height;j++) row_pointers[j]=(png_bytep *)(data+(j*(*width)*bytes));
+    /* Put image data into the PNG info structure. */
+    bytes = *nbits / 8;
+    row_pointers = malloc((*height) * sizeof(png_bytep));
+    for (j = 0; j < *height; j++)
```
AlexanderRichert-NOAA, line 147: There should be a space between `*` and `height`

AlexanderRichert-NOAA:
Are you using any automatic formatting tools? And if not would it be worth looking into?

src/enc_png.c
edwardhartnett: No because that's a pointer dereference.

edwardhartnett:
I would be up for using an autoformatter for fortran code.

Currently I use emacs which has a lot of build-in tools for this. BUt I'm trying out VSCode and I don't know yet what that provides...

edwardhartnett (approved):

PR #383, opened 2023-05-12
More testing for addlocal.f, and convert to addlocal.F90

edwardhartnett:
This brings testing for addlocal to 100%, and converts it to F90.


AlexanderRichert-NOAA (approved):
Builds/tests successfully on local machine. Confirmed 100% coverage for addlocal.F90.

Issue #379, opened 2023-05-12
Convert code to .F90


edwardhartnett:
Convert all code and test files to .F90, with free-format Fortran.

edwardhartnett:
OK, only pack_gp.f and reduce.f are still in F77 format. These files present special problems, and we have no testing for them. So I will close these issues and leave these two files as .f.

PR #384, opened 2023-05-12
converted g2_gbytesc, gribcreate, gribend to F90

edwardhartnett:
Part of #379 
Part of #191

These code files are already 100% covered in testing.


AlexanderRichert-NOAA (approved):
Code builds, tests pass, coverage for 3 files is still 100%.

Hang-Lei-NOAA (approved):

PR #391, opened 2023-05-15
Turn on and fix test for getgb2p(), convert it to F90

### LINKED ISSUE #380: Write test for GETGB2P
Currently there is no testing for this code.

edwardhartnett:
Fixes #380 
Part of #329

There was a test for getgb2p(), but it had been commented out. There was a problem that occured sometimes when make -j test was run, because both the _4 and _d version of the test were using the same LU.


.github/workflows/developer.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         mkdir build
         doxygen --version
         cd build
-        cmake .. -DENABLE_DOCS=On -DJasper_ROOT=~/Jasper -DCMAKE_PREFIX_PATH="~/bacio;~/w3emc" -DCMAKE_Fortran_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fsanitize=address" -DCMAKE_C_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fsanitize=address" -DFTP_TEST_FILES=ON -DCMAKE_BUILD_TYPE=Debug 
+        cmake .. -DENABLE_DOCS=On -DJasper_ROOT=~/Jasper -DCMAKE_PREFIX_PATH="~/bacio;~/w3emc" -DCMAKE_Fortran_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fno-omit-frame-pointer -fsanitize=address" -DCMAKE_C_FLAGS="-g -fprofile-abs-path -fprofile-arcs -ftest-coverage -O0 -Wall -fno-omit-frame-pointer -fsanitize=address" -DFTP_TEST_FILES=ON -DCMAKE_BUILD_TYPE=Debug
```
AlexanderRichert-NOAA, line 95: Is this related to #380 or #329?

edwardhartnett: The -fno-omit-frame-pointer causes more info to be printed when the address sanitizer finds a problem.

We should be using it in every developer.yml file, so if you are having some santizer failure, and you see it is not there, go ahead and add it.

It causes the error message to contain line numbers for the stack trace. Very helpful!

AlexanderRichert-NOAA (approved):
Builds and tests successfully on local machine

PR #413, opened 2023-05-25
fix memory leak in getgb2p()

### LINKED ISSUE #412: memory leak in getgb2p()
In getgb2p() we have:

`  save cbuf, nlen, nnum
`

Then we later allocate memory to cbuf in either getg2i() or getg2ir().

We never free this memory

Also note, from the GNU fortran manual:

`Note that variables which are explicitly initialized in declarations or in DATA statements automatically acquire the SAVE attribute.
`

So what happens is this:

```
  ! Determine whether index buffer needs to be initialized.
  irgi = 0
  if (lugi .gt. 0 .and. lugi .ne. lux) then
     call getg2i(lugi, cbuf, nlen, nnum, irgi)
     lux = lugi
  elseif (lugi .le. 0 .and. lugb .ne. lux) then
     mskp = 0
     call getg2ir(lugb, msk1, msk2, mskp, cbuf, nlen, nnum, nmess, irgi)
     lux = lugb
  endif

```  

Since lux and cbuf are both SAVE variables, the first time this is called, lux is set to the value of either lugb or lugi (the data or index file).

The next time the subroutine is called, lux is set so getg2i()/getg2ir() is not called and the value of cbuf from the previous call is used.

Seems like this idea cannot work. For example what happens if I call getgb2p(), then close the file and open a different one with the same LU, then call getgb2p() again. It will be using index information from the previous file.

edwardhartnett:
Fixes #412 

There are problems with how getgb2p() tries to optimize itself.

It uses SAVE to retain the value of a buffer of index data between calls. This saves the time reading the index info, either from the file or an index file. The second time the function is called, the index data does not have to be reread.

However, if the user changes the file the index info for subsequent calls to getgb2p() will be wrong. So this optimization will only work if only one file  is opened by the library. We have no way of guaranteeing that.

Also, the SAVE prevents us from releasing the memory, so it always leaks. 

To fix all this, I took out the optimization. No SAVE is used and the index info is reread from the file every time getgb2p() is called. I don't think this will add noticeable delay, it's pretty quick.


AlexanderRichert-NOAA (approved):
Test builds and passes on local machine

Issue #411, opened 2023-05-24
update CI


edwardhartnett:
Some changes to the CI would be helpful.

This stops the extra build of the branch in every PR:

```
on:
  push:
    branches:
    - develop
  pull_request:
    branches:
    - develop

```
It's good to have the name of the job be the same as the name of the workflow.

Under strategy:

`     fail-fast: true
`

Also now we get gcovr from apt-get instead of using pip. This seems to resolve the breaking gcovr problems we have had on some repos but not others.

Also gcc_docs_test-coverage.yml is a relic from a previous round of CI development, which can be removed. Everything it does (and more) is now done by developer.yml.

edwardhartnett:
Also the CI can build and run grib_util to make sure its tests also pass. I will add a new workflow for that.

edwardhartnett:
Also need to add TEST_FILE_DIR option to cmake to cache test files, just like in g2c.

edwardhartnett:
OK, this has all be done and the CI is looking pretty good for now.

PR #416, opened 2023-05-25
adding grib_utils workflow

edwardhartnett:
Adding grib_utils workflow...

Part of #411


edwardhartnett:
This is now in PR #418

PR #419, opened 2023-05-26
added TEST_FILE_DIR option to CMake, cached test data in the developer build

edwardhartnett:
Added TEST_FILE_DIR option, so that test files can be cached. Added caching for the test files to the developer workflow.

Part of #411


Hang-Lei-NOAA (approved):
I carefully checked that there is no ftp access/outreach in the code. This is good. The operation standard requested that no ftp, wget, git outreach in the code. We may just be careful when adding tests.

PR #442, opened 2023-06-14
updated README

### LINKED ISSUE #436: update README to document dependencies on bacio and w3emc
Hi,
I am installing the upp package on my local machine, and the first and second pass failed because of missing libraries although all libraries from the emc stack were installed.

going over the different steps once more, i noted that the  bacio library has a statement that the g2 library is using the bacio library. This to me reads like that there is a dependency on the bacio library. this dependency is not mentioned in the installation instructions. 

That suggests a confusion, meaning the upp installation suggests to install the g2 library first, (as it is on the top of the to-be-installed libraries) first, and bacio is listed in the middle. 

If there is a dependency on bacio (wihch would be consistent with the description of the bacio installation guide) could you add this to the dependency list? 

thanks
stephen

edwardhartnett:
Fixes https://github.com/NOAA-EMC/NCEPLIBS-g2/issues/436

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,8 @@
 
 # NCEPLIBS G2 Library
 
-Utilities for coding/decoding GRIB2 messages. This library contains
-Fortran 90 decoder/encoder routines for GRIB edition 2, as well as
-indexing/searching utility routines. This is part of the
+This library contains
+Fortran 90 decoder/encoder routines for GRIB edition 2. This is part of the
 [NCEPLIBS](https://github.com/NOAA-EMC/NCEPLIBS) project.
 
 For more detailed documentation see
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,17 +28,21 @@ Hartnett](mailto:edward.hartnett@noaa.gov)
...
 The following libraries are required:
-- libpng
-- libjasper
+
+This package requires the following third party libraries:
+- [Jasper](http://www.ece.uvic.ca/~mdadams/jasper/)
+- [libpng](http://www.libpng.org/pub/png/libpng.html)
+- [libz](http://www.gzip.org/zlib/)
+
+This package requires the folling NCEPLIBS libraries:
```
AlexanderRichert-NOAA: folling->following

README.md
```diff
@@ -29,17 +28,21 @@ Hartnett](mailto:edward.hartnett@noaa.gov)
 ## Prerequisites
 
 The following libraries are required:
-- libpng
-- libjasper
+
+This package requires the following third party libraries:
+- [Jasper](http://www.ece.uvic.ca/~mdadams/jasper/)
+- [libpng](http://www.libpng.org/pub/png/libpng.html)
+- [libz](http://www.gzip.org/zlib/)
```
AlexanderRichert-NOAA: I think this link should be https://www.zlib.net/ (current one doesn't work)

AlexanderRichert-NOAA: and I would rename libz->zlib

README.md
edwardhartnett: Fixed.

README.md
edwardhartnett: Done.

Hang-Lei-NOAA (approved):

Issue #425, opened 2023-05-28
complete testing for pdstemplates


edwardhartnett:
Lots of untested code here.

PR #444, opened 2023-06-15
Testing extendable templates in pdstemplates.F90

kingofnart:
Fixing #425 

Adding calls to extpdstemplate() for each extendable template. Need to check map length and map entries after calls.

### DIFF
--- a/tests/test_pdstemplates.F90
+++ b/tests/test_pdstemplates.F90
@@ -10,9 +10,7 @@ program test_pdstemplates
   integer :: idx
   integer :: nummap
   integer :: iret, i
-  integer, dimension(15) :: map_comp, list
-  integer, dimension(MAXLEN) :: map
-  !integer, dimension(MAXLEN) :: map1
+  integer, dimension(MAXLEN) :: map, map_comp, list
   logical :: needext
   integer :: pdtlen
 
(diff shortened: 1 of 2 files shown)


tests/test_pdstemplates.F90
```diff
@@ -45,22 +44,162 @@ program test_pdstemplates
...
+  print *, 'Template without map extension (needext = false)'
+  ! -- returns without doing anything
   call extpdstemplate(0, list, nummap, map)
-  if (nummap .ne. 0) stop 14
+  print *, 'Testing the extendable tempaltes'
+  ! ---- still need to test output i.e. nummap and map ----
+  ! Template number 3
+  call getpdstemplate(3, nummap, list, needext, iret)
+  if (iret .ne. 0 .or. nummap .ne. 31 .or. .not. needext) stop 14
+  call extpdstemplate(3, list, nummap, map)
```
edwardhartnett, line 150: This is good but you also need to test that the values returned in map and list are correct.

I suggest you add a bunch of variables which contain the lists you expect to get. Then check the lists you get against them, value by value.

kingofnart: I believe that some of the code won't run in extpdstemplate because in some cases it needs a value gotten from the map returned by getpdstemplate to be greater than 1 but the values checked seem to be 1 and so no extension takes place. I'm not sure if the values in the map aren't right or if it's the index being checked but some of the templates with extensions aren't being extended.

edwardhartnett: OK, let's get these tests merged and then I'll take a look...

edwardhartnett (approved):

Issue #424, opened 2023-05-28
test params_ecmwf functions with bad inputs


edwardhartnett:
We check how the functions in module params_ecmwf work with correct input, but we need to also check what happens when we try to look up values that don't exist.

Issue #361, opened 2023-05-07
Remove Program History Logs


edwardhartnett:
After a long consultation and examination of the documentation for NCEPLIBS-bufr, we decided to remove program history logs from our documentation.

These logs look like this:

!>    PROGRAM HISTORY LOG:
!>    - 2000-06-21 Stephen Gilbert Initial development.
!>    - 2011-10-24 Boi Vuong Added variable rmin4 for 4 byte float.

These don't actually add any information that is helpful, which is not already present elsewhere.

The initial development date and author are already listed for every file, subprogram, and module in the doxygen docs.

Other history log notes refer to changes in the code made by some programmer at some time. This is not particularly useful information, even to developers. It's totally useless to users.

All such information is currently managed through release notes and github history. All changes from the pre-github era are now so old as to be useless to list.

In fact, the program history section encourages the bad habit of writing documentation which is of no real use to anyone. It's a waste of time to maintain such lists.

So we removed them from NCEPLIBS-bufr and the resulting documentation is clearer, shorter, and easier to maintain. Programmers making changes to the code must integrate their documentation changes with the existing documentation, if necessary.

Since all PRs now have associate issues (like this one) there is no longer a need for program history logs in the code.

edwardhartnett:
This has been done.

PR #448, opened 2023-06-16
removed program log and added testing from params_ecmwf.f

AidanHartnett:
part of #424 and #361


tests/test_params_ecmwf.F90
```diff
@@ -15,6 +15,10 @@ program test_params_ecmwf
   if (g2disc .ne. 255 .or. g2cat .ne. 255 .or. g2num .ne. 255) stop 2
   call param_ecmwf_g1_to_g2(52, 1, g2disc, g2cat, g2num)
   if (g2disc .ne. 0 .or. g2cat .ne. 1 .or. g2num .ne. 1) stop 3
+  call param_ecmwf_g1_to_g2(-52, -1, g2disc, g2cat, g2num)
+  if (g1val .le. 0 .or. g1ver .le. 0) stop 5
```
edwardhartnett, line 19: You need to check the values of g2disc, g2cat, g2num after each call to param_ecmwf_g1_to_g2().

PR #451, opened 2023-06-16
More testing on pdstemplate.F90

kingofnart:
Setting some list entries to be greater than 1. 

Part of #425

### DIFF
--- a/tests/test_pdstemplates.F90
+++ b/tests/test_pdstemplates.F90
@@ -215,10 +215,7 @@ program test_pdstemplates
   do i = 1, nummap
     if (list(i) .ne. exp_map8(i)) stop 64
   end do
-  call extpdstemplate(8, list, nummap, map)
-  if (list(22) .le. 1) then
-    list(22) = 2
-  end if
+  list(22) = 2
   call extpdstemplate(8, list, nummap, map)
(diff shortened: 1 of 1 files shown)


tests/test_pdstemplates.F90
```diff
@@ -142,158 +184,302 @@ program test_pdstemplates
...
+    if (map(i) .ne. exp_extmap13(i)) stop 115
+  end do
   ! Template number 14
   call getpdstemplate(14, nummap, list, needext, iret)
   if (iret .ne. 0 .or. nummap .ne. 44 .or. .not. needext) stop 22
   do i = 1, nummap
     if (list(i) .ne. exp_map14(i)) stop 70
   end do
   call extpdstemplate(14, list, nummap, map)
+  if (list(37) .le. 1) then
```
edwardhartnett: Not sure why you have the if statements here. Seems like you could just set list(37) = 2, without caring what it was previously set to.

kingofnart: I guess I was thinking that if the maps ever changed so that the value being checked was something like 4 I wouldn't want to change it but that's unlikely to happen I suppose. I'll remove the if statements and the extra calls to extpdstemplate.

edwardhartnett (approved):

Issue #375, opened 2023-05-12
Test for gettemplates.f


edwardhartnett:
This code is completely untested.

PR #454, opened 2023-06-20
Adding testing for gettemplates.f

kingofnart:
Adding a test_gettemplates file in test directory.

Part of #375


edwardhartnett:
This broke because GitHub had some networking problems just when it ran. Push a change to the code to make the tests run again...

kingofnart:
I was able to get the code coverage up a little but there's still 4 error return codes I can't get to run. First I can't get unpack3 to return with error which I believe traces back to g2_gbytec needing to return a invalid template number which has to be encoded in the message. There is a similar challenge with unpack4 where the template number needs to be invalid to return error but the template number is gotten from g2_gbytec which must extract it from the input message. Also in the g2_gbytec function I need to be able to change the length of the message as I believe the message can't have both a section 3 and a section 4 to be able to get to the end of the message and read the '7777' end string so I've been running into seg faults when I shorten the message as the length still returns (from g2_gbytec) as 269. I was able to get the early '7777' error return since the program got to the end string before it seg faulted but the size of the message was still wrong. Checking if the message format is grib 2 is also gotten from g2_gbytec.

tests/test_gettemplates.F90
```diff
@@ -0,0 +1,502 @@
...
+  & char(k2), char(k2), char(k2), char(k2), char(k3),         &
+  & char(k3), char(k3), char(k3), char(k4), char(k4),         &
+  & char(k4), char(k4),char(k5),char(k5), char(k5),           &
+  & char(k5), char(0), char(k6), char(k6), char(k6),          &
+  & char(k6), char(k7), char(k7), char(k7), char(k7),         &
+  & char(k8), char(k8), char(k8), char(k8), char(k8),         &
+  & char(k8), char(k8), char(k8), char(0), char(80),          &
+  & char(k6), char(k6), char(k6), char(k6), char(k6),         &
+  & char(k6), char(k6), char(k6), char(k9), char(k9),         &
+  & char(k9), char(k9), char(0), char(0), char(0), char(0),  &
```
edwardhartnett, line 239: This is fine but take a look in test_gribcreate.F90 and you'll see a different way.

Only store one valid GRIB2 message, but then use code like this to change a value and run a test:

  ! Change the first byte of the message, then try to add grid - will
  ! not work.
  old_val = cgrib(1)
  cgrib(1) = char(0)
  call addgrid(cgrib, lcgrib, igds, igdstmpl, igdstmplen, &
       ideflist, idefnum, ierr)
  if (ierr .ne. 1) stop 50
  cgrib(1) = old_val

Note that I change the first byte, then try an addgrid() call, and it fails. THen I reset the first byte to its correct value, ready for the next test.

Using this technique, you should be able to reach most of the other error checking code...

edwardhartnett: ???

Issue #421, opened 2023-05-26
Finish testing for gf_unpack2()


edwardhartnett:
In gf_unpack2() we have this untested code:

```
      if ( isecnum.ne.2 ) then
         ierr=6
         print *,'gf_unpack2: Not Section 2 data. '
         return
      endif

```
This could be tested by constructing a bad grib message, with the wrong section number.

PR #456, opened 2023-06-24
Testing for gf_unpack2

kingofnart:
Adding a test file and improving the code coverage for gf_unpack2.

Part of #421

### DIFF
--- a/src/drstemplates.F90
+++ b/src/drstemplates.F90
@@ -211,17 +211,15 @@ contains
     ! No implemented DRS templates need extensions.
     if (.not. templates(index)%needext) return
 
-    ! Uncomment the code below if we ever implement a template that
-    ! needs extension.
-    ! nummap = templates(index)%mapdrslen
-    ! map(1 : nummap) = templates(index)%mapdrs(1 : nummap)
+    nummap = templates(index)%mapdrslen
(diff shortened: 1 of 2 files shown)


tests/test_gf_unpack2.F90
```diff
@@ -0,0 +1,107 @@
...
+    iofst = 49
+    allocate(csec2(10))
+    print *, 'Testging gf_unpack2'
+    ! Normal grib message with offset to section 3
+    call gf_unpack2(fgrib, fgrib_len, iofst, lensec2, csec2, ierr)
+    if (ierr .ne. 6) stop 1
+    iofst = 38
+    call gf_unpack2(fgrib, fgrib_len, iofst, lensec2, csec2, ierr)
+    print *, 'ierr: ', ierr
+    deallocate(csec2)
```
edwardhartnett: This is good to check the error handling code that was not being checked.

But can you also include a call to gf_unpack2 which works, and check the results are correct?

I know this is already done elsewhere, but in future, when wanting to look for a simple test, test_gf_unpack2.F90 is where we will look...

edwardhartnett: Can you include an additional test, and then we can get this merged?

edwardhartnett: ???

Issue #426, opened 2023-05-28
complete testing of mkieee.f


edwardhartnett:

edwardhartnett:
OK, we have improved but still have more to go. From the code coverage report:

![image](https://github.com/NOAA-EMC/NCEPLIBS-g2/assets/38856240/5936d2ff-bb0c-4ad2-b0d9-6cf81c38f0a2)

So seems like we need to try a value of exactly zero...

PR #460, opened 2023-07-04
Adding testing to test_mkieee.F90

AidanHartnett:
part of #426

### DIFF
--- a/tests/test_mkieee.F90
+++ b/tests/test_mkieee.F90
@@ -11,7 +11,7 @@ program test_mkieee
 #if KIND == 4
   integer, parameter :: num = 4
   real :: rieee(num)
-  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000 /)
+  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000 -2.3300000 -1.6700000 /)
   real :: b(num)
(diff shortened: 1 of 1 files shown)


tests/test_mkieee.F90
```diff
@@ -29,6 +29,11 @@ program test_mkieee
      print *, i, a(i), b(i)
      if (abs(a(i) - b(i)) .gt. .00001) stop 20
   end do
+
+  call mkieee(a, rieee, num)
+
+  if (atemp .lt. 0.0 .and. n .ne. -1) stop 30
```
edwardhartnett: atemp and n are not involved here.

Instead you should be checking the value returned in array rieee.

AidanHartnett: I'm trying to test the case where n=-1. The value is already checked early in the test file.

edwardhartnett: I don't even understand how this compiled, because atemp is not defined and we have implicit none. So how come we're not getting a variable not found error?

edwardhartnett: What you need to do is extend array a by adding some negative numbers, and then make sure we get the correct result from mkieee, which means check the values returned in rieee.

Basically, mkieee takes array a, with length num, and then fills array rieee with the ieee versions of those numbers.

PR #466, opened 2023-07-07
Adding testing to test_mkieee.F90

AidanHartnett:
part of #426


tests/test_mkieee.F90
```diff
@@ -9,9 +9,9 @@ program test_mkieee
   implicit none
 
 #if KIND == 4
-  integer, parameter :: num = 4
+  integer, parameter :: num = 5
   real :: rieee(num)
-  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000 /)
+  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000 -4.3000000 /)
```
edwardhartnett: You need a comma.

edwardhartnett: OK, there is another bug in this test.

Where it says:

`#if KIND == 4
`
Change it to:

`#ifdef KIND_4
`

Currently none of the code in the test is ever being compiled. ;-)

That's how you can forget a comma and not get a compile error.

edwardhartnett: See #467

tests/test_mkieee.F90
```diff
@@ -29,6 +29,13 @@ program test_mkieee
      print *, i, a(i), b(i)
      if (abs(a(i) - b(i)) .gt. .00001) stop 20
   end do
+
+  call mkieee(a, rieee, num)
+
+  do i = -1, num
```
edwardhartnett: Take lines 36-38 out.

The value of the output will be checked on line 30.

src/mkieee.f
```diff
@@ -59,7 +59,7 @@ subroutine mkieee(a,rieee,num)
               n = n - 1
            enddo
         endif
-!        n=floor(alog(atemp)/alog2)
+        n=floor(alog(atemp)/alog2)
```
edwardhartnett: Is this change unintentional? You should not need to change the code, just the tests.

edwardhartnett: If you can remove this change we can merge the rest of the PR and get these tests into use...

AidanHartnett: ok sounds good

edwardhartnett (approved):

edwardhartnett:
I will merge this after #480

PR #473, opened 2023-07-08
removed IF (ASSOCIATED(CBUF)) because CBUF is undefined

### LINKED ISSUE #472: in ixgb2() and other subroutines, we check whether a pointer is associated, and it is undefined, causing a memory error in valgrind
We get this error:

```
==14601== Conditional jump or move depends on uninitialised value(s)
==14601==    at 0x115D49: ixgb2_ (ixgb2.f:73)
==14601==    by 0x10B61C: MAIN__ (test_ixgb2.F90:60)
==14601==    by 0x10B931: main (test_ixgb2.F90:7)

```
This is because of this line of code:

`      IF (ASSOCIATED(CBUF)) NULLIFY(CBUF)
`

CBUF is passed as an argument to the function. In the caller (for example getg2ir()) does not initialize the buffer, so the pointer is undefined. Calling ASSOCIATED() on an undefined pointer is an error.

As it says in the documentation of ASSOCIATED(PTR [, TGT]): "The status of neither PTR nor TGT can be undefined."

Some further explanation:
* https://www.cs.rpi.edu/~szymansk/OOF90/bugs.html#5
* https://gcc.gnu.org/onlinedocs/gcc-4.3.6/gfortran/ASSOCIATED.html

This problem also occurs in getg2i(), and getg2ir().

edwardhartnett:
Fixes #472


Hang-Lei-NOAA (approved):
When this is removed. Should we add allocate and deallocate to prevent the CBUF memory leak?

edwardhartnett:
@Hang-Lei-NOAA the allocate/deallocate are already there in the code.

The only effect of this change is that the cbuf pointer is not checked for associated() before it has been nullified. Doing so is actually a violation of the Fortran standard, which is incredibly awkward, but true,

The final solution to this problem can take one of two forms:
1 - Change the Fortran code to use allocatable arrays instead of pointers. This makes memory leaks impossible.
2 - Change the Fortran library to use this functionality in the C library, and remove it from Fortran.

Probably 2 is the answer we will go with eventually. But for now, let's get the existing codebase running without leaks, and well tested, so that when we do start to swap in C function calls, we will know that the results are correct and memory does not leak.

PR #482, opened 2023-07-11
Adding test cases to test_mkieee.F90

AidanHartnett:
part of #460


tests/test_mkieee.F90
```diff
@@ -9,7 +9,7 @@ program test_mkieee
 #ifdef KIND_4
   integer, parameter :: num = 7
   real :: rieee(num)
-  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000, -4.3000000, 0.3300000, -400.0000000 /)
```
edwardhartnett, line 12: Instead of taking out the test for -400.00000 and adding one for 0, just add one for 0.

In other words, leave the existing test value and add a new value to the test array.

edwardhartnett: Can you make this change and then we can merge this?

AidanHartnett:
I'm trying to do that but visual studio code keeps crashing everytime I try
to commit

edwardhartnett (approved):

Issue #36, opened 2020-11-03
Need to make correction for opening file larger than 2 GB


edwardhartnett:
Need to make correction for opening file larger than 2 GB.

BoiVuong-NOAA:
Hi Ed, 
I am unable to solve in G2 library for opening file larger than 2 GB. Please re-assign to someone in nceplibs team.

edwardhartnett:
The test code should create a file larger than 2 GB and then try to open it.

Hang-Lei-NOAA:
This has been addressed on v3.4.2 and corrected on v3.4.3.
Okay to close.

kgerheiser:
Hang this was not related to v3.4.2 which was for opening file units > 100.

kgerheiser:
What I've gathered is that there is an integer that's overflowing somewhere and it needs to be changed to 64-bit. Gfortran has the flag `-ftrapv` which is supposed to catch integer overflow. g2 could be compiled with that and then run a > 2GB file through g2 to see where the overflow occurs.

kgerheiser:
https://github.com/NOAA-EMC/NCEPLIBS-g2/blob/e184a5d375507840b3935cc1affb84e57d8fdb33/src/skgb.f#L32

I was able to get DEGRIB2 to work after changing the integers here to int64. There's more, and anything that deals with byte counts needs to be made int64.

MatthewPyle-NOAA:
I'm also running into issues reading > 2 GB GRIB2 files, so happy to see that this issue is being addressed.  Is there a timeline for version 3.5.0?

GeorgeGayno-NOAA:
I am also having problems with > 2GB files.  That function must work to support this ufs_utils project - https://github.com/ufs-community/UFS_UTILS/issues/660.  That is a high priority project according to @EricRogers-NOAA.

edwardhartnett:
@kgerheiser  so in file getg2ir.F90 did you make these variables:
lugb, iseek, msk1, lskip, lgrib

into INTEGER(KIND=8)? Or did you turn all the integers into KIND=8?

kgerheiser:
I'm trying this again with @GeorgeGayno-NOAA's test case.

GeorgeGayno-NOAA:
> I'm trying this again with @GeorgeGayno-NOAA's test case.

Let me add some diagnostic print so you can see where it breaks.

kgerheiser:
[...]

PR #492, opened 2023-07-17
adding large file testing

edwardhartnett:
Adding large file testing.

This will only work on my machine until more of my test files are uploaded to the FTP site.

Part of #36 

I would like to get this merged to aid test development, although the tests of the FV3 file are not very well developed yet. I will be expanding them over the next few days.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,27 @@ set(lib_name ${PROJECT_NAME})
 
 include(GNUInstallDirs)
 
+# Handle user build options.
+option(ENABLE_DOCS "Enable generation of doxygen-based documentation." OFF)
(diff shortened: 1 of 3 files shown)


Hang-Lei-NOAA (commented):
Where is test data? if the size is larger than 2 GB, it cannot be loaded to github.

edwardhartnett:
@Hang-Lei-NOAA the test data is at the moment on my hard drive. Eventually it will be on the FTP site. If the user selects the correct options, they will be able to download and run the test, but, in practice, it is something only we will do, in one of our CI systems, with caching so the data is not constantly downloaded.

Hang-Lei-NOAA (approved):

PR #493, opened 2023-07-18
adding skgb8() to handle GRIB2 files > 2 GB

### LINKED ISSUE #36: Need to make correction for opening file larger than 2 GB

edwardhartnett:
Fixes #36 

Handling large files.

In this PR I add function skgb8(), which is like skgb(), except it takes integer*8 arguments. This should allow the _4 build of the library handle files > 2 GB.

More testing will follow, but this is ready to merge and works for files < 2 GB just as skgb() did.


AlexanderRichert-NOAA (approved):
Might be worth leaving the issue open until the testing is completed, or create a new issue? Otherwise no concerns here.

PR #496, opened 2023-07-19
More testing of gf_unpack2

kingofnart:
Showing memory leak in gf_unpack2 when giving offset to section 2 data.

Part of #421


edwardhartnett:
OK, so you have some leaks in your test code:

```
Direct leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x7f06789a9867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x556346e5c126 in test_gf_unpack2 /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:113
    #2 0x556346e5d839 in main /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:173
    #3 0x7f0678172d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

Direct leak of 37 byte(s) in 1 object(s) allocated from:
    #0 0x7f06789a9867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x556346e7920c in gf_unpack2_ /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/src/gf_unpack2.f:49
    #2 0x556346e5d173 in test_gf_unpack2 /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:137
    #3 0x556346e5d839 in main /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:173
    #4 0x7f0678172d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

Direct leak of 37 byte(s) in 1 object(s) allocated from:
    #0 0x7f06789a9867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x556346e5c93b in test_gf_unpack2 /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:125
    #2 0x556346e5d839 in main /home/runner/work/NCEPLIBS-g2/NCEPLIBS-g2/g2/tests/test_gf_unpack2.F90:173
    #3 0x7f0678172d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58


```
So if we look at line 113, we see:

`    allocate(csec2(100))
`
and sure enough I see later in the code:

`    deallocate(csec2)
`
But you have also allocated the memory on lines 113 and lines 133. Each time you allocate, you must deallocate. If you allocate 3 times and deallocate once, then two allocations will be leaking. ;-)

But what I also see is that in gf_unpack2(), this memory is also allocated, on line 49:

`      allocate(csec2(lencsec2),stat=istat)
`
When I look at the documentation for this parameter, I see:

`!> @param[out] csec2 Pointer to a character*1 array containing local use data.
`
Which is not very clear. So, to get your test working, remove all your allocate() statement, but leave the deallocate statement. And leave the nullify statement. 

The documentation needs to be updated, so change the documentation in gf_unpack2.f to be:

`!> @param[out] csec2 Pointer to a character*1 array containing local use data. Caller should pass in an unassociated pointer, and must free memory allocated by this subroutine by calling deallocate on this pointer.
`

This is a good example of how adding testing also causes the documentation to improve.

kingofnart:
When I call gf_unpack2 without first allocating csec2 (to at least size 8) then I get a seg fault which points to line 36 in gf_unpack2 which is nullify(csec2). I also get a segfault when I allocate and deallocate csec2 before the function call.

edwardhartnett:
OK, this is a weird one.

Because there is a pointer in the parameter list, we need to have an interface for the subroutine. I thought that Fortran gave an error and told us that, but instead of just does something weird with memory.

So to start with, insert the following lines after you declare variables but before any program statements:

```
    interface
       subroutine gf_unpack2(cgrib, lcgrib, iofst, lencsec2, csec2, ierr)
         character(len = 1), intent(in) :: cgrib(lcgrib)
         integer, intent(in) :: lcgrib
         integer, intent(inout) :: iofst
         integer, intent(out) :: lencsec2
         integer, intent(out) :: ierr
         character(len = 1), pointer, dimension(:) :: csec2
       end subroutine gf_unpack2
    end interface
```
    

Then your program works better. You have one unneeded deallocate at the end of your test.

kingofnart:
I added the interface and it works great now, thanks. I was still only able to get one of the two error cases checked and I'm not sure I'll be able to get the second one since the `allocate(csec2(lencsec2),stat=istat)` statement on line 51 in gf_unpack2 needs to fail. Since the subroutine calls `nullify(csec2)` before later allocating the array, `csec2` should in theory never be allocated a second time at line 51 and so that error should never happen? Not too sure about that but that's my impression. I still am able to get errors though by calling gf_unpack2 with specific offsets. I found all of the offsets that give section 2 and their respective section lengths:
| Offset | Length |
| --- | --- |
| 24 | 1107296251 |
| 81 | -5 |
| 117 | 564133883 |
| 129 | 37 |
| 142 | 344123 |
| 169 | 1048571 |
| 185 | 507 |
| 257 | 336860155 |
Any offset with a large section length ends up passing the allocate check but seg faults on `csec2(1:lencsec2)=cgrib(ipos:ipos+lencsec2-1)` (line 58 of gf_unpack2). I'm also still confused why offset 129 gives the most reasonable section length and offset 38 doesn't read as section 2 (which is the index of the start of section 2 in the grib message).

edwardhartnett:
OK, making progress but those section lengths look wrong...

kingofnart:
I agree they definitely don't look right, by my count the offset should be 38 and the section length should be 11. I also checked the lengths by running the code that's in gf_unpack2 in the test program without making any gf_unpack2 calls and I still get the same section lengths returned by g2_gbytesc.

kingofnart:
It looks like I didn't read the documentation well enough because you supply the message length in bytes and the offset in bits. If the offset byte is 37 (bit 296) then the section length (in octets) is 6. This makes sense because it agrees with the length given by grib_info, although I still don't know why the offset isn't byte 38 or bit 304. Either way supplying an offset that gives a section number of 2 and a large section length still seg faults the program.

edwardhartnett:
OK, can you comment out the test code that causes the segfault? Then I can get the rest merged, and that will make it easier for me to help get the code work.ing...

edwardhartnett (approved):

PR #508, opened 2023-07-24
converted params_ecmwf.f and getgb2s.f to F90 and added implicit none

edwardhartnett:
Part of #379 
Part of #191
Converted code to F90.


Hang-Lei-NOAA (approved):
Although it is in f90 standard, the uppercase letters make getgb2s.F90 look like still be f77.

PR #512, opened 2023-07-24
added huge test case to test_mkieee.F90

AidanHartnett:
part of #426


edwardhartnett:
You  are getting this error:

```
   12 |   real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000, -4.3000000, 0.3300000, 0.0000000, -400.000000, 101029284739847594/)
      |                                                                                                                                    1
Error: Line truncated at (1) [-Werror=line-truncation]

```
The problem is that F90 only allows lines of a certain length (120 IIRC). So this line of code is getting cut off.

Use the continuation character to break the line in two. To break a line, insert an ampersand and a new line, like this:

```
 real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000, -4.3000000, 0.3300000, 0.0000000, &
          -400.000000, 101029284739847594/)

```

edwardhartnett:
You could also remove the trailing 0s from the constants. In other words, make 0.3300000 into 0.33. That would make the line shorter.

edwardhartnett:
Any progress here?

tests/test_mkieee.F90
```diff
@@ -9,7 +9,7 @@ program test_mkieee
 #ifdef KIND_4
   integer, parameter :: num = 9
   real :: rieee(num)
-  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000, -4.3000000, 0.3300000, 0.0000000, -400.000000, 101029284739847594/)
+  real :: a(num) = (/ 4.3000000, 5.6000000, 1.6700000, 2.3300000, -4.3000000, 0.33, 0.0, -400.0, 101029284739847594/)
```
edwardhartnett: You can take the trailing 0s away from all the numbers in the array.

edwardhartnett:
OK, the problem with the grib_utils workflow is caused by something we did on grib-utils today, not by your code.

Let me investigate...

edwardhartnett (approved):

PR #514, opened 2023-07-30
converted gf_unpack6 to F90, adding implicit none

edwardhartnett:
converted gf_unpack6 to F90, adding implicit none.

This code file is fully tested.

Part of #191 
Part of #379


AlexanderRichert-NOAA (approved):
Build and tests successfully with gcc 8 & 12 and intel. I don't see any warnings for the new F90 file.

Issue #378, opened 2023-05-12
Test other types of dimensions in getdim.f


edwardhartnett:
Currently only lat/lon is being tested.

Some messages will have to be constructed using the other types, before getdim can be tested on them.

PR #519, opened 2023-08-01
Adding tests for other dimensions in getdim.f

kingofnart:
Still need to test with template number 30 (Lambert Conformal) and the default case when the template number isn't recognized. I ran into some memory errors there were a value being used was uninitialized.

Part of #378


edwardhartnett (approved):
We will get this merged and take a look at what is going on in the test...

Issue #377, opened 2023-05-12
Improve testing for addfield.f


edwardhartnett:
There are a lot of conditions in this code that are not being tested.

edwardhartnett:
With recent merges much of the easy testing has been done, and remaining testing involves some grib complexities.

1. bitmaps - some GRIB messages have them, but I have no examples to work with (yet).
2. different types of compression, currently only simpack() is tested.

Of these, 2 looks easiest to tackle. Getting JPEG and PNG compression working in tests would be very helpful. This would mean changing the DRS template array (idrstmpl in the parameter list of addfield()). By setting the DRS template to the appropriate size and values, different compression types can be specified.

PR #520, opened 2023-08-02
More testing on addfield.f

kingofnart:
Part of #377 

Added a test for errror cases 3 and 6 in addfield.f


edwardhartnett:
Best hope for debugging memory errors is firstly to comment out as much of the test code as you need to so that the test passes.

Then uncomment the smallest piece of test code which causes the memory error and debug that...

edwardhartnett:
Please comment out whatever code is causing the memory issues so we can get the rest of the test merged and start working on the memory problems. It's hard for me to help when this is on a branch of your fork.

kingofnart:
The issue I'm having is that all the tests pass on my machine. It looks like the last error had to do with the simpack call on line 289 of addfield.f but I can't reproduce the error locally.

edwardhartnett (approved):

edwardhartnett:
I'm going to merge what you;'ve got here and work on it in a separate PR to get the rest of the test code working...

edwardhartnett:
OK, there are two different ways we run memory tests in the CI system, and either can be done on your machine and should reproduce the problem.

The one that is failing is the address sanitizer. In order to use the address sanitizer we add the following flags to our C flags and our Fortran flags:
-fno-omit-frame-pointer -fsanitize=address

You can reproduce this in your own cmake build by adding to the cmake command line:

cmake -DCMAKE_C_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -DCMAKE_Fortran_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -DCMAKE_BUILD_TYPE=Debug etc.

So give it a try...

edwardhartnett:
Also for examples of how to run the memchecks or anything else look in file .github/workfows/developer.yml.

edwardhartnett:
OK, there was only one problem with this code, which was that the real array was declared real*4. Once the *4 was removed the test works.

The reason is that we  build with -d option as well, which has default real as 8-byte, unless you explicitly set the size. So when you set it to *4 it is *4 even on *8 builds, and breaks. When we leave it unset, it is *4 on *4 builds, and *8 on *8 builds.

See #530 for the working test code. Thanks for this test it is very useful.

kingofnart:
Oh that's right I remember disabling ASAN to use valgrind for debugging but I never added the -fsanitize=address flags back. Ok, that makes sense why it was only the _d test that failed then.

PR #522, opened 2023-08-04
figuring out problems in test_gf_unpack2

edwardhartnett:
Part of #421 

Working on test of gf_unpack2().

### DIFF
--- a/tests/test_gf_unpack2.F90
+++ b/tests/test_gf_unpack2.F90
@@ -4,197 +4,192 @@
 !
 ! Andrew King 6/23/23
 
- program test_gf_unpack2
-    
-    implicit none
+program test_gf_unpack2
 
-    character, dimension(269) :: fgrib
-    integer :: k1, k2, k3, k4, k5, k6, k7, k8, k9, k10
-    integer :: n1, n2, n3, nan, fgrib_len
-    integer :: iofst, lencsec2, lensec, ierr, i
-    integer :: isecnum, numlocal, numfields
-    integer :: listsec0(3), listsec1(13), maxvals(7)
-    character(len=1),pointer,dimension(:) :: csec2
+  implicit none
 
-    interface
-        subroutine gf_unpack2(cgrib, lcgrib, iofst, lencsec2, csec2, ierr)
-        character(len = 1), intent(in) :: cgrib(lcgrib)
-        integer, intent(in) :: lcgrib
-        integer, intent(inout) :: iofst
-        integer, intent(out) :: lencsec2
-        integer, intent(out) :: ierr
-        character(len = 1), pointer, dimension(:) :: csec2
-        end subroutine gf_unpack2
-    end interface
+  character, dimension(269) :: fgrib
+  integer :: k1, k2, k3, k4, k5, k6, k7, k8, k9, k10
+  integer :: n1, n2, n3, nan, fgrib_len
+  integer :: iofst, lencsec2, lensec, ierr, i
+  integer :: isecnum, numlocal, numfields
+  integer :: listsec0(3), listsec1(13), maxvals(7)
+  character(len=1),pointer,dimension(:) :: csec2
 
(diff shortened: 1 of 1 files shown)


tests/test_gf_unpack2.F90
```diff
@@ -4,197 +4,192 @@
...
+  call gf_unpack2(fgrib, fgrib_len, iofst, lencsec2, csec2, ierr)
+  if (ierr .ne. 0 .or. iofst .ne. 384 .or. lencsec2 .ne. 6) stop 1
+  deallocate(csec2)
+
+  ! Offset not to section 2 data, should error without allocating csec2
+  iofst = 1
+  call gf_unpack2(fgrib, fgrib_len, iofst, lencsec2, csec2, ierr)
+  if (ierr .ne. 6) stop 2
+
+  ! Offset to section 2 data with already allocated csec2 array
```
edwardhartnett: This seems like an attempt to make the allocate fail on line 53 of gf_unpack2.f. Unfortunately that is not what happens. Instead, fortran happily allocates more memory and just loses the previously allocated memory, unless you kept a separate copy of the pointer. So there is no way to make an allocation fail in testing, Some code will necessarily remain untested as a result.

kingofnart: That's what I was attempting here. I also noticed that there is a nullify statement before the allocate statement in gf_unpack2 anyway so there's really no chance the allocate statement could error I don't think.

tests/test_gf_unpack2.F90
```diff
@@ -4,197 +4,192 @@
...
+  ! print *, 'Offset: ', i, 'length: ', lencsec2
+  ! iofst = 465
+  ! i = iofst
+  ! call gf_unpack2(fgrib, fgrib_len, iofst, lencsec2, csec2, ierr)
+  ! deallocate(csec2)
+  ! print *, 'Offset: ', i, 'length: ', lencsec2
+
+  ! print *, ''//NEW_LINE('A')//'Calling gf_unpack with offset 24 ...'
+  ! iofst = 24
+  ! call gf_unpack2(fgrib, fgrib_len, iofst, lencsec2, csec2, ierr)
```
kingofnart: I still have a question here, is the user expected to know the correct offset and it's unlikely that the wrong offset would be used? If you choose the offset carefully you can get the program to think it's unpacking section 2 but the length is bogus and causes a seg fault. Not sure if it's significant but I thought I would bring it up.

tests/test_gf_unpack2.F90
```diff
@@ -3,255 +3,128 @@
...
-    & char(8), char(0), char(0), char(1), char(0), char(1),    &
-    & char(7), char(n1), char(11), char(15), char(10),          &
-    & char(10), char(10), char(0), char(1),                      &
-        ! section 2
-    & char(0), char(0), char(0), char(11), char(2),             &
-    & char(1), "H", "K", "T", "A", "R",                             &
-        ! section 3
-    & char(0), char(0), char(0), char(81), char(3), char(0),   &
-    & char(0), char(0), char(0), char(k10), char(0), char(0),  &
-    & char(0), char(30), char(1), char(0), char(k1), char(k1), &
```
edwardhartnett, line 66: OK, this is incorrect in several places. THe problem is that the char() function returns a character*1, which is 8 bytes. So no value greater than 255 makes sense as the argument.

edwardhartnett (commented):
cdd

tests/test_gf_unpack2.F90
edwardhartnett: Correct, the caller just has to know the correct offset for section 2. Calling with an incorrect offset which coincidentally falls on a patch of data which looks like the start of section 2 will do nutty things.

edwardhartnett:
Test still needs some work but this is an improvement...

Hang-Lei-NOAA (approved):

PR #526, opened 2023-08-07
Created new test file for mkieee.f using chatGPT

AidanHartnett:
part of #426


tests/test_mkieee_ChatGPT.F90
```diff
@@ -0,0 +1,54 @@
+program ieee_array_large_conversion_test
```
edwardhartnett: Add a comment explaining this file, with your name as author, "using ChatGPT". Also put the exact prompt you used to generate this code.

edwardhartnett:
Wow ChatGPT did a lot better than I thought it would!!

edwardhartnett:
This test is currently not running. You need to add it to the tests in the tests/CMakeLists.txt file. Just do for this file what is already done for test_mkieee.

edwardhartnett:
image

edwardhartnett:
OK, now that I look at this, I'm kind of surprised that the mac was the only one to fail!

One complexity here is that we are running tests for a _4 build of the library and a _d build. The _4 build has regular 4-byte reals, but the _d build has 8-bytes as the default real. This is a bad idea but we are stuck with it for now.

But clearly, from the code, mkieee() is only intended to be used with 4-byte reals. It's not clear whether rdieee() will work with 8-byte reals, but it might. However, some of the code in addfield() certainly looks like it will fail _d builds, because it seems to be passing 8-byte reals into mkieee(), which clearly only accepts 4-byte reals.

OK, so change your test so the whole test is ifdef'd away unless KIND_4 is set. Like this:

program whatever
#ifdef KIND_4

... all the test code

#endif
end program whatever

This will resolve the problems with _d builds, but macos is also failing with _4 builds. However, let's resolve the _d first and then circle back to the _4 problem.

edwardhartnett:
Please add the ifdef as described and lets see if we can get this test working...

edwardhartnett:
Comment out this test in the tests/CMakeLists.txt file, so that the test is not being run.

That will allow me to merge it and then check it out easily to work on it...

tests/test_mkieee_ChatGPT.F90
```diff
@@ -0,0 +1,60 @@
+! Test the mkieee and rdieee functions in NCEPLIBS-g2 written by ChatGPT using the prompt "write a fortran test program that tests an array of positive and negative values which includes a very large number and zero for the mkieee()/rdieee() functions of the NCEPLIBS-g2 library"
+!
+! Aidan Hartnett
+
+!program ieee_array_large_conversion_test
```
edwardhartnett, line 5: Don't comment out the lines:

program ieee_array_large_conversion_test

and

end program ieee_array_large_conversion_test

PR #535, opened 2023-08-22
fixed error handling in getg2ir()

### LINKED ISSUE #400: getg2ir.f() calls stop for a memory problem, instead of returning error code
In getg2ir.f we have:

```
           DEALLOCATE(CBUFTMP,STAT=ISTAT)
           IF (ISTAT.NE.0) THEN
             PRINT *,' deallocating cbuftmp ... ',istat
             stop 99
           ENDIF

```

Instead we need to return an error.

edwardhartnett:
Fixes #400 

Fix an error handling problem in getg2ir().

### DIFF
--- a/src/getg2ir.F90
+++ b/src/getg2ir.F90
@@ -31,6 +31,7 @@
 !> - 0 No error.
 !> - 1 Not enough memory available to hold full index buffer.
 !> - 2 Not enough memory to allocate initial index buffer.
+!> - 3 Error deallocating memory.
 !>
 !> @author Mark Iredell @date 1995-10-31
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA:
@edwardhartnett off the top of your head, is there a test case that covers this scenario?

edwardhartnett:
No, sadly this cannot be tested. To test it would be necessary to make deallocate fail. In languages like python, this can easily be accomplished by a technique called "mocking", which allows you to say "make the first deallocate() call in this test return failure". Using mocking you can easily achieve 100% code coverage, which is great.

But I don't know of any way to use mocking in Fortran...

AlexanderRichert-NOAA (approved):

PR #536, opened 2023-08-23
fixed workflow, converted some files to F90

### LINKED ISSUE #523: grib_util not using different jasper versions
There's a typo in the grib_utils.yml CI workflow, specifically, 'jasper-vesion' should be 'jasper-version'. Based on https://github.com/NOAA-EMC/NCEPLIBS-g2/actions/runs/5761765347/job/15620228333, it appears to be using the same cached copy of jasper for both jasper versions in the matrix (because the version part of the cache key is missing).

edwardhartnett:
Fixes #523 
Part of #379

Fixes workflow to use jasper versions.


AlexanderRichert-NOAA (approved):
Compiles and tests successfully with gcc11 on personal machine.

PR #539, opened 2023-08-27
converted getgb2rp to F90

edwardhartnett:
Part of #379 

Converted getgb2rp to F90.


AlexanderRichert-NOAA (approved):
Builds and tests successfully with gcc11 on personal machine.

PR #542, opened 2023-08-29
large file test

edwardhartnett:
Testing the handling of large files (> 2GB).

Part of #36


edwardhartnett:
OK, this add a 6 GB file to the FTP testing. I actually can't believe this worked because I thought GitHub would time out downloading it, or object to the size of the data cache afterwards. But it did not!

If this causes future CI runs to be really slow we can turn off the large file testing.

edwardhartnett:
@AlexanderRichert-NOAA or @Hang-Lei-NOAA please review.

Hang-Lei-NOAA (approved):
That's really good. I was thinking that large file transfer must through the git lfs. You did it through ftp. We could turn it off after testing, if anyone feel that it is too big or NCO concerns the data security.

edwardhartnett:
Well as long as we store the data file in the cache with GitHub, it should work OK.

NCO does not have to run the FTP-based tests if they don't want to. They are not run by default, you have to specify a cmake build option to include FTP testing.

PR #567, opened 2023-09-26
test that gf_finalize() can be called multiple times with no harm

### LINKED ISSUE #551: test that gf_finalize() can be called multiple times without harm
test that gf_finalize() can be called multiple times without harm

edwardhartnett:
Fixes #551


tests/test_getgb2.F90
```diff
@@ -125,6 +125,10 @@ program test_getgb2
   call gf_finalize(iret)
   if (iret .ne. 0) stop 5
 
+  ! Call finalize again, should do nothing.
+  call gf_finalize(iret)
+  if (iret .ne. 0) stop 5
```
AlexanderRichert-NOAA, line 130: Should this be a different stop #?

AlexanderRichert-NOAA (approved):

tests/test_getgb2.F90
edwardhartnett: OK

Issue #557, opened 2023-09-11
Fix shared library build


AlexanderRichert-NOAA:
BUILD_SHARED_LIBS doesn't seem to be having any effect. When we fix this, we should also add checks to the CI to make sure the correct library is being built.

edwardhartnett:
OK, I've got a fix for this, but it turns out that w3emc does not build with shared libs, and that needs to be fixed first.

PR #572, opened 2023-10-02
check shared library build with a make install

edwardhartnett:
Part of #557 

Make sure shared library build is working.


edwardhartnett:
Got this working elsewhere.

PR #579, opened 2023-10-06
updating package file

edwardhartnett:


spack/package.py
```diff
@@ -36,7 +37,8 @@ class G2(CMakePackage):
     )
     variant("w3emc", default=True, description="Enable GRIB1 through w3emc", when="@3.4.6:")
 
-    depends_on("jasper@:2.0.32")
+    depends_on("jasper@:2.0.32", when="@:3.4.7")
+    depends_on("jasper", when="@3.4.8:")
```
edwardhartnett: @AlexanderRichert-NOAA did I get this right? I am trying to say that jasper-2.0.32 is required up to g2-3.4.7, and starting with g2-3.4.8, any version of jasper can be used.

However, judging by the spack output, it's still using jasper-2.0.32.

AlexanderRichert-NOAA: Yes that looks right. You can take out the `when="@3.4.8:"` in that second line just as a matter of convention, but it shouldn't be hurting anything either. The CI output for this PR looks correct (I'm seeing jasper 3.0.3), and the behavior is correct when I test on my PC: I can build g2@3.4.8 with whatever jasper version I want, but if I try to build 3.4.7 with jasper 3, it throws an error.

Hang-Lei-NOAA (approved):

PR #581, opened 2023-11-14
documentation improvements

edwardhartnett:
Part of #580

### DIFF
--- a/.github/workflows/Intel.yml
+++ b/.github/workflows/Intel.yml
@@ -44,8 +44,7 @@ jobs:
         sudo apt-get update
         sudo apt-get install intel-oneapi-dev-utilities intel-oneapi-mpi-devel intel-oneapi-openmp intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
         echo "source /opt/intel/oneapi/setvars.sh" >> ~/.bash_profile
-        which ifort
-        which ifx
+        which ifort || which ifx
       
     - name: checkout-bacio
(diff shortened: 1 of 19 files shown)


Hang-Lei-NOAA (approved):

edwardhartnett:
@AlexanderRichert-NOAA any idea why ifort is no longer working? I didn't change anything...

AlexanderRichert-NOAA:
In the place where it's currently failing (the "which" line), setvars.sh hasn't actually been sourced yet, so I'm assuming that's why it's not finding it. As far the previous error of "Could not find compiler set in environment variable FC" I'm not sure. Maybe try putting the "which" right before the cmake call?

AlexanderRichert-NOAA:
It's a problem with Intel. I just did my own installation on Ubuntu 22 following the workflow steps, and it installs the same versions and everything. Version 2023.2.2 doesn't provide ifort (note that the fortran package version is 2024). I think what we'll need to do is specify the version we want, probably 2023.2.1. As far as I understand it, icc/icpc are not and will not be part of the 2024 release.

edwardhartnett:
@AlexanderRichert-NOAA I don't quite understand from your response what should be done...

AlexanderRichert-NOAA:
Yeah sorry-- try specifying the compiler versions, i.e. "intel-oneapi-compiler-fortran-**2023.2.1** intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic-**2023.2.1**" and see if that does the trick.

Issue #600, opened 2024-02-04
g2_sbytesc() does not work for arrays of 64-bit ints


edwardhartnett:
g2_sbytesc() does not work for arrays of 64-bit ints.

See test at end of test_gbytec.F90:

```
    testing g2_sbytec() with 64-bit int array of size 4...
                    1                    1                    1                    1
  1 00
  2 00
  3 00
  4 00
  5 00
  6 00
  7 00
  8 01
  9 00
 10 00
 11 00
 12 00
 13 00
 14 00
 15 00
 16 00
 17 00
 18 00
 19 00
 20 00
 21 00
 22 00
 23 00
 24 01
 25 00
 26 00
 27 00
 28 00
 29 00
 30 00
 31 00
 32 00
    now unpack into 4 64-bit ints with g2_gbytesc()...
                    1                    1                    0                    0
```

PR #602, opened 2024-02-05
more testing for gbytec subroutines plus some minor doxygen fixes

edwardhartnett:
Part of #550 
Part of #600

### DIFF
--- a/tests/test_gbytec.F90
+++ b/tests/test_gbytec.F90
@@ -11,6 +11,9 @@ program test_gbytec
   character*1 :: out5(5)
   character*1 :: out8(8)
   character*1 :: out10(10)
+  character*1 :: out16(16)
+  character*1 :: out32(32)
+  integer :: in4(4), in1(1)
   integer, parameter :: n = 1
   integer :: in(n)
   real :: r_in(n)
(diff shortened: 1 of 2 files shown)


tests/test_gbytec.F90
```diff
@@ -172,10 +195,73 @@ program test_gbytec
...
+  call g2_sbytesc(out32, in84, iskip, 64, 0, 4)
+  do i = 1, 32
+     print '(i3, x, z2.2)', i, out32(i)
+  end do
+
+  print *, '   now unpack into 4 64-bit ints with g2_gbytesc()...'
+  do i = 1, 4
+     in84_1(i) = 0
+  end do
+  call g2_gbytesc(out32, in84_1, iskip, 64, 0, 4)
```
AlexanderRichert-NOAA, line 262: Are there values that need to be checked here, or is the idea (for now anyway) to just ensure that the routine doesn't fail?

edwardhartnett: I'm going to submit a follow-on with more testing and fix for the int*8 issue.

AlexanderRichert-NOAA (approved):

PR #603, opened 2024-02-05
changes to support large index files

edwardhartnett:
Part of #550 

This PR contains changes to start to support the index files for > 2 GB.

In this PR I switch to using bareadl() instead of baread() for the ixgb2() subroutine. To maintain backward compatibility I introduce a new ix2gb2() subroutine, which will accept the index version as a parameter. Calling ixgb2() will call ix2gb2() with an index version of 1 (i.e. legacy behavior). But ix2gb2() can be used on files larger than 2 GB.

I would like to merge these changes incrementally, rather than as one giant PR. The changes in this PR have no effect on functionality at the moment, but allow for the next version of the index to be supported in future PRs.

### DIFF
--- a/src/g2index.F90
+++ b/src/g2index.F90
@@ -646,11 +646,11 @@ END SUBROUTINE GETGB2S
 !> - byte kk + 1-  ll the data representation section (drs)
 !> - byte ll + 1-ll + 6 first 6 bytes of the bit map section (bms)
 !>
-!> @param[in] lugb Unit of the unblocked GRIB file. Must
+!> @param lugb Unit of the unblocked GRIB file. Must
 !> be opened by [baopen() or baopenr()]
 !> (https://noaa-emc.github.io/NCEPLIBS-bacio/).
-!> @param[in] lskip Number of bytes to skip before GRIB message.
-!> @param[in] lgrib Number of bytes in GRIB message. When subroutine is
+!> @param lskip Number of bytes to skip before GRIB message.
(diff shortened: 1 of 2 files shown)


src/g2index.F90
```diff
@@ -265,25 +265,25 @@ END SUBROUTINE GETG2I
...
-       INTEGER, INTENT(IN) :: LUGB, LSKIP, LGRIB
-       CHARACTER(LEN = 1), POINTER, DIMENSION(:) :: CBUF
-       INTEGER, INTENT(OUT) :: NUMFLD, MLEN, IRET
-     END SUBROUTINE IXGB2
-  END INTERFACE
+  interface      ! required for cbuf pointer
+     subroutine ixgb2(lugb, lskip, lgrib, cbuf, numfld, mlen, iret)
+       integer :: lugb, lskip, lgrib
+       character(len = 1), pointer, dimension(:) :: cbuf
+       integer :: numfld, mlen, iret
```
AlexanderRichert-NOAA, line 284: Is there a reason for removing the intent's here? `lugb` for instance is documented under `ixgb2()` as having intent 'in' but I don't see that being set anywhere.

edwardhartnett: The intents are not in the code, and are a mistake in the documentation.

I would add them to the code but I'm not sure how that would affect in existing code. So instead I'm removing them from the documentation.

AlexanderRichert-NOAA (approved):

PR #606, opened 2024-02-07
Now support index for GRIB2 files > 2 GB

edwardhartnett:
Part of #550 

This adds more code to support an index for files > 2 GB. There are a lot of changes necessary to support this one additional 64-bit value in the index file. ;-)

More changes coming in the next PR, but this one can be merged. All changes are made in a backward compatible way, so that all existing code continues to work just as it always has.

To get the index files for files > 2 GB, new functions must be called, wtih an index version argument.


edwardhartnett:
@AlexanderRichert-NOAA or @Hang-Lei-NOAA please review...

edwardhartnett:
Never mind, this PR superseded by #607

PR #610, opened 2024-02-13
Rest of index v2 changes

edwardhartnett:
Part of #550 

These are all the changes needed to get index version 2 working.

This includes testing, but more testing will be coming in the next few PRs, including testing with some very large sample files provided by users.


AlexanderRichert-NOAA (approved):
A few param descriptions in src/getgb2r.F90 could use capitalizing, but otherwise looks good

Hang-Lei-NOAA (approved):

PR #613, opened 2024-02-15
more testing for intmath

### LINKED ISSUE #598: recover test code in test_intmath.F90
We have some legacy test code in test_intmath.F90 (commented out). Can it be recovered? If not, it should be removed.

edwardhartnett:
Fixes #598 

Recovered old intmath test code.


tests/test_intmath.F90
```diff
@@ -34,40 +41,28 @@ program test_intmath
...
+        temp = log(real(max(ival + one, one), kind = 16)) / alog2
+     endif
+     iret = i1log2(ival)
+     if (iret /= ceiling(temp) .or. ival == 0 .or. ival == check) then
+        if (iret /= ceiling(temp)) then
+           print 10, ival, ival, iret, ceiling(temp),temp
+        else
+           print 20, ival, ival, iret, ceiling(temp),temp
+        endif
+     endif
```
AlexanderRichert-NOAA: How does/would this fail? Is there supposed to be a 'stop' somewhere in here?

edwardhartnett: Good catch! Stand by...

edwardhartnett: Done.

Hang-Lei-NOAA (approved):

PR #621, opened 2024-02-23
adding test

edwardhartnett:
Part of #550 
More testing.

### DIFF
--- a/tests/test_create_index_seaice.F90
+++ b/tests/test_create_index_seaice.F90
@@ -1,0 +1,175 @@
+! This is a test program for NCEPLIBS-g2.
+!
+! This program tests index file functionality with g2_create_index()
+! for one of the files optionally downloaded via FTP during testing:
+! seaice.t00z.grb.grib2
+!
+! Ed Hartnett 2/23/24
+program test_create_index_seaice
+  use grib_mod
+  implicit none
+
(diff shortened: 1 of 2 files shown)


tests/test_create_index_seaice.F90
```diff
@@ -0,0 +1,175 @@
...
+
+        ! Free memory.
+        call gf_free(gfld)
+        print *, '   OK!'
+     end do
+
+     ! Clean up.
+     deallocate(cbuf)
+     call gf_finalize(iret)
+     if (iret .ne. 0) stop 200
```
AlexanderRichert-NOAA, line 171: Would it make sense to change this to something other than `stop 200` since that number is already used?  Otherwise looks good.

AlexanderRichert-NOAA (approved):

tests/test_create_index_seaice.F90
edwardhartnett: Sure, but I really only track that down if I start having trouble with the test.

That is, if the test starts failing, and I need them, then I will fix up the stop numbers. But a repeated number is not a bug because it is still enough to cause the test to fail.

So you don't have to worry about the stop numbers in test codes when you inspect them. Just make sure a non-zero stop happens when there is a problem.

PR #626, opened 2024-03-05
adding subroutines g2_gbytec1()/g2_sbytec1()

### LINKED ISSUE #625: Add versions of g2_gbytec()/g2_sbytec() which take scalar integer arguments
We often call g2_gbytec()/g2_sbytec() with a scalar integer, but these functions want integer arrays.

Since we do this so often in the code, it seems better to add subroutines g2_gbytec1()/g2_sbytec1(), which take a scalar integer argument.

This is necessary for #624.

edwardhartnett:
Fixes #625


edwardhartnett:
@Hang-Lei-NOAA or @AlexanderRichert-NOAA please review.

Hang-Lei-NOAA (approved):

PR #630, opened 2024-03-10
fixed warnings about unused parameter in subroutine reduce() (and pack_gp())

### LINKED ISSUE #586: unused argument warning in reduce.f
We have:

```
/home/ed/NCEPLIBS-g2/src/reduce.f:42:30:

   42 |       SUBROUTINE REDUCE(KFILDO,JMIN,JMAX,LBIT,NOV,LX,NDG,IBIT,JBIT,KBIT,
      |                              1
Warning: Unused dummy argument ‘kfildo’ at (1) [-Wunused-dummy-argument]

```

When I look in reduce.f I see:

!> @param[in] KFILDO unit number for output/print file.

and later, all the KFILDO code is commented out.

```
C***                  WRITE(KFILDO,135)L,NOV(L),NOVL,M,J,IBXX2(J)
C*** 135              FORMAT(/' AT 135--L,NOV(L),NOVL,M,J,IBXX2(J)',6I10)               

```

The subroutine reduce() is called by the pack_gp() subroutine which is some pretty hairy code. It also has a KFILDO parameter, also all commented out; the only remaining use is in the reduce() call, so if I take it out of there, it will just move this warning up to pack_pg() and also change the API.

edwardhartnett:
Fixes #586

Removed unused parameter from pack_gp() and reduce().


Hang-Lei-NOAA (commented):
the subroutine reduce still use the kfildo variable. 
This may also need to be decreased.

edwardhartnett:
@Hang-Lei-NOAA good catch. I have removed it...

Hang-Lei-NOAA (approved):

Issue #651, opened 2024-04-22
We need a getgb2p2() for > 2 GB files


edwardhartnett:
We need a getgb2p2() for > 2 GB files.

I missed this when doing the recent > 2GB work...

edwardhartnett:
With help from @AlexanderRichert-NOAA , this is now complete. Thanks Alex!

PR #652, opened 2024-04-22
more > 2 GB work

edwardhartnett:
Part of #651

### DIFF
--- a/src/g2getgb2.F90
+++ b/src/g2getgb2.F90
@@ -420,7 +420,8 @@ subroutine getgb2l2(lugb, idxver, cindex, gfld, iret)
   endif
 end subroutine getgb2l2
 
-!> Find and extract a GRIB2 message from a file.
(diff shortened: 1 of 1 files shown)


src/g2getgb2.F90
```diff
@@ -570,14 +590,111 @@ end subroutine getgb2rp
...
 
-  ! Free cbuf memory allocated in getg2i/getg2ir().
+  ! Free cbuf memory allocated in getg2i/getg2i2r().
   if (associated(cbuf)) deallocate(cbuf)
   
   call gf_free(gfld)
+end subroutine getgb2p2
+
+!> Find and extract a GRIB2 message from a file, using index version
+!> 1. Use getgb2p2() to use either index version 1 or version 2.
```
AlexanderRichert-NOAA, line 604: I'm a little confused by the documentation-- Does this subroutine do just index version 1 or does it handle both?

Issue #653, opened 2024-04-22
Index format 2 will not handle all > 2 GB files


edwardhartnett:
Unfortunately, based on a conversation with @webisu I have learned that further modifications are required for index format 2 to work on all large GRIB2 files.

Large GRIB2 files will have sections < 2 GB. But since they can be 2 GB, indexes to a section within a message must be 8-byte ints. Currently we use 4-byte ints. We need to change to 8 byte ints.

edwardhartnett:
@AlysonStahl-NOAA and @AlexanderRichert-NOAA this will be a problem for tocgrib2 as well.

edwardhartnett:
OK, this has been a ton of work, but the index file now uses 8-byte ints everywhere that it should.

The only thing left for this issue is to inspect the code for cases where I still use 4-byte ints internally. All these values need to be 8-byte everywhere in the code.

PR #660, opened 2024-05-06
converted some index values to 8-byte

edwardhartnett:
Part of #653 

Converting some 4-byte values in the index to 8-byte values in the index.


AlexanderRichert-NOAA:
Testing this with tocgrib2 and the test file from the RRFS devs, if I use `getgb2p`, it segfaults (on the line where it sets the first of the four 7's), and if I use `getgb2p2`, it doesn't seem to resolve the missing 7777 problem.

edwardhartnett:
It would be helpful if you could approve this and I can merge it, as I continue to solve the problems with > 2GB that I recently learned of from Wesley.

I'm not working on the 7777 problem yet, but the first step will be an automated test that demonstrates the problem. Do we have that yet?

Is the 7777 problem in tocgrib2 or g2? Do we know? If it is in g2, we should have a g2 test that demonstrates it without involving tocgrib2.

Is there an issue for the 7777 problem? If not, can you create one?

AlexanderRichert-NOAA (approved):

AlexanderRichert-NOAA:
Yes the issue is in g2, I just created an issue for it: https://github.com/NOAA-EMC/NCEPLIBS-g2/issues/666
@edwardhartnett if you're able to keep on with the coding side I can work on setting up a CTest test using the RRFS file.

PR #662, opened 2024-05-06
Fix index 1 vs. 2 bug in g2getgb2.F90

AlexanderRichert-NOAA:
This fixes the problem with extract=.true. for RRFS files with tocgrib2. This change doesn't seem to break the unit tests (namely, test_files).

Part of #653


edwardhartnett:
OK, this is a good fix, thanks! I'm going to merge it...

edwardhartnett (approved):

edwardhartnett:
@AlexanderRichert-NOAA also can you provide a test for this change?

PR #687, opened 2024-05-16
separating memcheck into separate CI workflow

### LINKED ISSUE #686: need to break apart the developer workflow
THe developer workflow does two builds, one with asan and one with valgrind checking.

But these two builds try to share the same data cache and that is not working.

SO I will split them into two workflows and see if that helps.

edwardhartnett:
Fixes #686

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -18,7 +18,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  developer:
+  memcheck:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA:
What's the issue with the caches? And is the idea that they'll have separate ones?

edwardhartnett:
@AlexanderRichert-NOAA or @Hang-Lei-NOAA please review. The developer.yml workflow download of FTP files over and over is really slow.

AlexanderRichert-NOAA (approved):

PR #706, opened 2024-05-30
Combining grib_util with this repo

### LINKED ISSUE #296: combine grib_util with this repo
Seems like we can move the grib utilities in with g2.

We can add an option ENABLE_UTILITIES=OFF for those who want to build g2 without the utils.

Since grib_util depends on g2 already, this will mean the same software gets built, but with one less repo, build system, and dependency.

@Hang-Lei-NOAA @BoiVuong-NOAA @DusanJovic-NOAA comments welcome.

### LINKED ISSUE #332: move all grib_util utilities to the g2 library
@AlexanderRichert-NOAA @Hang-Lei-NOAA this will impact you, but hopefully make our lives easier.

Immediately after the next releases of g2 and grib_util, I plan to move all the grib_util code to the g2 repo (except wgrib).

Some of the grib_util code needs the w3emc library, and some is GRIB2 only. The ones that need w3emc will only be built if the g2 library is build with w3emc. So we will be able to build g2 without w3emc, and it would then include utilities that do not depend on w3emc.

The reason for this is to reduce our install, spack, and testing work. Instead of two packages, we will just have one to manage. And when working on utility bugs (as we are now with tocgrib2), it would be a lot easier to develop tests if the utility and g2 code were in the same repo.

The next release of grib_util will be the last one, and wgrib (which is the GRIB1 version of wgrib2) will only be available in grib_utils. If anyone still wants it, I can see about moving it to a subdirectory of the wgrib2 project. But it does not belong with the g2 code and the rest of grib_utils, which are just very light utilities that make use of g2.

Comments welcome...

AlysonStahl-NOAA:
Fixes #296 
Fixes NOAA-EMC/NCEPLIBS-grib_util#332


edwardhartnett:
The tests for utils should be in their own directory, test_utils. It should have it's own data subdirectory.

But we don't want duplicate data files. If the utils test needs to get data from tests/data that''s fine too.

AlysonStahl-NOAA:
@edwardhartnett @AlexanderRichert-NOAA 

I've done pretty much all I can to get most of the checks to pass. 

The tests for the utilities fail the memory checks in the develop workflow. Is this something we can work around for now and fix down the line or should we make sure the tests pass those checks before merging?

One of the Spack checks keep failing. I did make changes to the python script myself and I'm fairly new to Spack, so someone should probably double check my work there.

I know the grib_util workflow can probably be deleted, but I want to double check before I do that.

Also, I just need to update the docs, but I'm not sure how we want to structure it. Any thoughts?

AlexanderRichert-NOAA:
The Spack workflow failure is because it's compiling w3emc without some extra functions that are disabled by default. In the grib-util recipe, notice it depends on `w3emc +extradeps`, so we'll need the same thing here.

edwardhartnett:
The grib_utils CI run was the one that build grib_utils and ran its tests in order to check g2. Since we will now be building and testing them together, it's no longer needed and I will remove it...

edwardhartnett:
The grib_utils will fail memory tests. That's a known problem. For the purposes of this PR, just cause the developer build to not build the utilities at all. We will address this with some CI cleanup after @AlexanderRichert-NOAA merges his CI simplifying changes which will make it easier.

I will see if I can do that edit to the developer workflow...

.github/workflows/Linux_versions.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       matrix:
         jasper-version: [2.0.33, 3.0.5, 4.0.0]
         bacio-version: [2.4.1, 2.5.0, 2.6.0]
-        w3emc-version: [2.9.2, 2.9.3, 2.10.0]
+        w3emc-version: [2.10.0, 2.11.0]
```
edwardhartnett: Do w3emc-2.9.2 aand w3emc-2.9.3 no longer work with g2?

edwardhartnett:
OK, I turned off utils in the developer memory checking build. Also I removed the linux version expected builds for two versions of w3emc, but why?

@AlexanderRichert-NOAA can you help with the spack stuff? I tried to figure it out but ended up confused...

Once we get the CI problems resolved, I think we can merge...

.github/workflows/Linux_versions.yml
AlysonStahl-NOAA: Yes, the grib_util library required w3emc > 2.10.0. 

It just occurred to me thought that this particular workflow does not build the tools, so I can revert the changes to make sure older versions are still getting tested.

AlexanderRichert-NOAA:
Sure. I think I see the issue. I'll see if I can work out a fix and I'll submit it as a PR to @AlysonStahl-NOAA's branch.

AlexanderRichert-NOAA:
See https://github.com/AlysonStahl-NOAA/NCEPLIBS-g2/pull/1. It might be worth (at some point) adding BUILD_4/BUILD_D to the non-Spack unit tests under Linux_options.

AlysonStahl-NOAA:
> See [AlysonStahl-NOAA#1](https://github.com/AlysonStahl-NOAA/NCEPLIBS-g2/pull/1). It might be worth (at some point) adding BUILD_4/BUILD_D to the non-Spack unit tests under Linux_options.

Good catch, thanks for fixing that.

edwardhartnett:
@AlysonStahl-NOAA do you know what is happening with the failing build in CI?

Did we change the acceptable version of w3emc?

AlysonStahl-NOAA:
> @AlysonStahl-NOAA do you know what is happening with the failing build in CI?
> 
> Did we change the acceptable version of w3emc?

Yes I did change that. I changed it back to the previous requirement (2.9.0) if BUILD_UTILS is off. It will require 2.10.0 and higher if it's on.

AlexanderRichert-NOAA:
FWIW this looks pretty good to me-- the only thing I would add is you'll want to update ./cmake/PackageConfig.cmake.in to reflect dependencies, including adding OpenMP if enabled.

AlysonStahl-NOAA:
> FWIW this looks pretty good to me-- the only thing I would add is you'll want to update ./cmake/PackageConfig.cmake.in to reflect dependencies, including adding OpenMP if enabled.

Which other dependencies should be included? It looks like the only dependency it checks for currently is PNG and bacio was removed recently. It doesn't seem to check for Jasper or w3emc either.

AlexanderRichert-NOAA:
Are any of the changes as far as added find_package's, OpenMP, etc. related to the library itself, or is that all just for the utilities? If it's just for the utilities then I guess nothing needs to change there.

AlysonStahl-NOAA:
The find_package changes should be for the utilities only. In fact, I'll go ahead and only look for those packages if the utilities are enabled.

edwardhartnett:
Is this ready to merge?

AlysonStahl-NOAA:
> Is this ready to merge?

Yep.

edwardhartnett (approved):

Issue #328, opened 2022-11-21
expose some of the new g2c functions in the C library


edwardhartnett:
The C library contains some g2c functions, like g2c_open(), which I would like to  use in testing in NCEPLIBS-g2c. For that it would be good to have these functions exposed in the fortran API.

This is related to #23

edwardhartnett:
There is a test, test_g2cf.F90, which contains:
```
program test_g2cf
  use g2cf
  implicit none
  character (len = *), parameter :: fileName = "gdaswave.t00z.wcoast.0p16.f000.grib2"
  integer :: g2cid, num_msg
  integer(kind = 1) :: discipline
  integer(kind = 4) :: num_fields, num_local
  integer(kind = 2) :: center, subcenter
  integer(kind = 1) :: master_version, local_version
  integer :: ierr

  print *, 'Testing g2cf API...'
  ierr = g2cf_set_log_level(1)

  ! Open the test file.
  ierr = g2cf_open(fileName, 0, g2cid)
  if (ierr .ne. 0) stop 2

```

At one point I had this test code working, when I was prototyping all this last year, but now it is commented out in the cmake file, and there is no sign of a g2cf.F90 module file anywhere. So somehow that got lost. 

I will recreate it and get the test working again.

PR #714, opened 2024-06-12
adding g2cf.F90

edwardhartnett:
Part of #328


edwardhartnett:
OK, this must wait for #708 to be merged before it can proceed.

src/g2c_interface.F90
```diff
@@ -0,0 +1,36 @@
+! Contains a module with interfaces to the C functions in the g2c
+! library.
+!
+! Note: this file is excluded from the doxygen build (see Doxyfile.in)
+! because these functions are already documented in the g2c library.
```
AlexanderRichert-NOAA, line 5: Does/should the doxygen docs have some indication of which functions have Fortran interfaces?

src/g2cf.F90
```diff
@@ -0,0 +1,157 @@
...
+       cstring = string(1:nlen)
+    else ! append null char to trimmed string
+       cstring = string(1:nlen)//C_NULL_CHAR
+       nlen = nlen + 1
+    endif
+  end function addcnullchar
+
+  !> Open a GRIB2 file.
+  !>
+  !> @param path the path to the file
```
AlexanderRichert-NOAA: Did we decide to capitalize the doxygen descriptions of variables? Personally I would find it easier to read if either the first letter of descriptions were upper case, or if there were some separator like a hyphen.

src/g2cf.F90
```diff
@@ -0,0 +1,157 @@
...
+    integer :: ie1, ie2
+    
+    cmode = mode
+    cg2cid = 0
+    
+    ! check for c null character on path and add one if not present.
+    cdata_file = addcnullchar(data_file, ie1)
+    cindex_file = addcnullchar(index_file, ie2)
+    
+    ! call nc_create to create file
```
AlexanderRichert-NOAA: I'm assuming you don't mean nc_create :)

src/g2c_interface.F90
edwardhartnett: No users access everything through the fortran API and never call these functions directly.

src/g2cf.F90
edwardhartnett: Fixed, thanks.

Hang-Lei-NOAA (approved):

PR #715, opened 2024-06-12
Clean up utils directory

### LINKED ISSUE #712: cleanup utils directory
The utils directory has one file per subroutine, which leads to messy directories.

Instead, where reasonable, the util should contain it's subroutines within the one file. For example we have:

`  add_executable(degrib2 degrib2.F90 prlevel.F90 prvtime.F90)
`

The prlevel and prvtime subroutines can just be added to degrib2.F90, making this a one-file solution, much simpler.

This can be done for most of the utils. If the resulting file is going to be too big, then don't do this, but do it for all the utils with just a few files.

AlysonStahl-NOAA:
Fixes #712


edwardhartnett (approved):

edwardhartnett:
@AlysonStahl-NOAA sorry, I didn't notice this was still just a draft. Looks good though!

AlysonStahl-NOAA:
I think this should be done now. 

The file for cnvgrib would be too long if I moved all of it's subroutines into it, but I managed to condense it down to 4 total files.

The subroutines used by degrib2 are used by a test so I moved them all into one separate file. 

makwmo.f90 is used by two utilities, so I kept that separate as well.

PR #722, opened 2024-06-25
Finding differences in pds templates between g2 and g2c

AlysonStahl-NOAA:
Part of https://github.com/NOAA-EMC/NCEPLIBS-g2c/issues/383

### DIFF
--- a/src/pdstemplates.F90
+++ b/src/pdstemplates.F90
@@ -39,7 +39,7 @@
 module pdstemplates
 
   integer, parameter :: MAXLEN = 200 !< MAXLEN max length of entries
-  integer, parameter :: MAXTEMP = 43 !< MAXTEMP maximum number of templates
+  integer, parameter :: MAXTEMP = 47 !< MAXTEMP maximum number of templates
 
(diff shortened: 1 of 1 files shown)


AlysonStahl-NOAA:
@edwardhartnett I've found a few differences between the templates here and in g2c. I copied over the arrays from g2c just so I can keep track of where the differences are. I will change some back and fix the ones in g2c if they are incorrect over there, I just need to figured out which one is the correct one.

I understand what the numbers mean and that the negative values in the array indicate that a negative value is possible. How do I know which ones can be negative?

I am aware that the templates are defined here:
https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_sect4.shtml

edwardhartnett:
How many codes with differences are there?

AlysonStahl-NOAA:
> How many codes with differences are there?

There were 5 templates with differences in sign. There were also 4 templates that existed in g2c and not in g2. I added those missing templates.

In some cases g2 had the negative sign and g2c didn't. Should I assume the case with the negative sign is the correct one?

PR #12, opened 2020-06-08
CMake, CI, and Tests

kgerheiser:
Same old thing I've been doing.

Builds and runs:

https://github.com/kgerheiser/NCEPLIBS-g2/runs/751492737?check_suite_focus=true

PR #737, opened 2024-08-08
Fixed pds template #12

AlysonStahl-NOAA:
Part of https://github.com/NOAA-EMC/NCEPLIBS-g2c/issues/383


AlexanderRichert-NOAA (approved):
@AlysonStahl-NOAA is it correct to say these template values are all consistent with what's in g2c?

AlysonStahl-NOAA:
> @AlysonStahl-NOAA is it correct to say these template values are all consistent with what's in g2c?

Yep, more differences were found (#722), but I've been asked to keep the changes in separate PRs.

edwardhartnett:
How did you determine which was correct, g2 or g2c?

AlysonStahl-NOAA:
> How did you determine which was correct, g2 or g2c?


I'm double checking the [definitions here.](https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_temp4-12.shtml)
The first change from negative to positive was for Forecast Time, which from I understand from previous conversations can be negative. For the next change, the value was "Total number of data values missing in the statistical process" which I assumed would only ever be positive.

edwardhartnett:
I think it would be best to add an issue for each template that has to change, and in the issue, describe it as you did above, and link to the WMO documentation.

Then of course link that issue to the PR.

PR #738, opened 2024-08-12
Use g2c

AlexanderRichert-NOAA:
This PR is intended to add AEC functionality from g2c.

It needs testing, incl. unit testing, and I'm still not 100% on the C bindings, especially with respect to handling pointers.


edwardhartnett:
@AlexanderRichert-NOAA looks like this is the answer. Why didn't it work?

@AlysonStahl-NOAA this will be helpful...

PR #739, opened 2024-08-12
Use g2c for AEC

AlexanderRichert-NOAA:
(reopening PR since it won't let me reuse the old one)

This PR is intended to add AEC functionality from g2c.

It needs testing, incl. unit testing, and I'm still not 100% on the C bindings, especially with respect to handling pointers.


AlexanderRichert-NOAA:
@edwardhartnett @AlysonStahl-NOAA I pushed from the wrong local copy of the updated code, so please ignore the old PR I closed and use this one, which is basically the same changes except it also includes my attempt to define Fortran interfaces with C bindings.

AlexanderRichert-NOAA:
@edwardhartnett would it make sense to put the Fortran interfaces under g2c itself so developers (including ourselves) don't have to duplicate code? I'm making some progress in sorting out handling pointers, char's, etc. as far as the ISO C bindings go.

edwardhartnett:
Yes good idea.

edwardhartnett:
OK, the interfaces should be in g2 not g2c, since g2c does not and should not require a Fortran compiler.

Currently the interfaces are in src/g2c_interface.F90. Add whatever you need there. (Also that file is excluded from doxygen build.)

Can this pull request be closed? I am about to do a release...

AlexanderRichert-NOAA:
Noted. I believe that @AlysonStahl-NOAA (who I hope will correct me if I'm wrong) added aec support from g2c into g2 (with the f2c interface built into g2).

PR #843, opened 2026-04-29
g2getgb2.F90: fix sec6 index parsing

AlexanderRichert-NOAA:
This PR fixes https://github.com/NOAA-EMC/NCEPLIBS-grib_util/issues/358. The added unit testing fails with a segfault with v3.5.1.

Once merged into the new branch, develop-3.5.x, then we can drop a 3.5.2 release.


AlexanderRichert-NOAA:
I'm probably not going to try to keep debugging the CI, but the main thing is that the "memcheck" workflow succeeded, meaning that there was no segfault when sec 6 was deliberately overloaded in the test_getb2_? tests.

Hang-Lei-NOAA (approved):

